PR #5, opened 2016-04-11
spec: Remove scylla-server dependency

benoit-canet:
(To be merged after scylla "[PATCH] spec: Extract scylla.yaml and create metapackage" and the equivalent ubuntu patch)

Now that the scylla.yaml configuration file is in scylla-conf
we can remove the dependency on scylla-server.

Fixes #421

Signed-of-by: Benoît Canet <benoit@scylladb.com>


penberg:
@benoit-canet, please send the fix as a patch just like with other repositories.

PR #47, opened 2017-08-02
2454

benoit-canet:
Remove the extra properties checker from the tools so we can get rid on
the scylla.yaml to cassandra.yaml synchronisation script.

Second patch remove it.


tools/bin/cassandra.in.sh
```diff
@@ -46,25 +46,12 @@ fi
...
-    # cassandra options parser go boom.
-    # Also add attributes not present in scylla.yaml
-    # but required by cassandra.
-    `dirname $0`/filter_cassandra_attributes.py \
-            "$CASSANDRA_CONF/cassandra.yaml" \
-            "$TMPCONF/scylla.yaml" \
-            > "$TMPCONF/cassandra.yaml"
-    CASSANDRA_CONF=$TMPCONF
-    fi
+cp -a $SCYLLA_CONF/scylla.yaml $SCYLLA_CONF/cassandra.yaml
```
slivne: why do we need to copy ?

benoit-canet:
@slivne I think the tool diggest a cassandra.yaml file. Before those patches the python script was doing the "copy".

benoit-canet:
@slivne: Note there is extra logging infos at the begining of the ouput with the default scylla configuration file

tools/bin/cassandra.in.sh
slivne: @benoit-canet - so why not change that - e.g. force the code to use the scylla.yaml instead

benoit-canet:
@slivne: iterated

benoit-canet:
@penberg Hi pekka I am asking you to review this :)

benoit-canet:
@penberg: can you review this please ?

slivne:
@penberg ping

penberg:
There are patches on the mailing lists, which I commented on.

PR #68, opened 2018-09-17
Origin 3.11

elcallio:
Update scylla-tools-java to origin 3.11


elcallio:
Note that the respective patch sets for dtest and ccm are required to run dtest (due to string matching diffs).

penberg:
I applied those too. Thanks, @elcallio!

PR #89, opened 2019-03-24
nodetool/TableHistograms: Print an error when no keyspace/table found

amnonh:
Instead of throwing an exception, print an error that the keyspace/table
is not found.

See scylladb/scylla-jmx#55

Signed-off-by: Amnon Heiman <amnon@scylladb.com>

### DIFF
--- a/src/java/org/apache/cassandra/tools/nodetool/TableHistograms.java
+++ b/src/java/org/apache/cassandra/tools/nodetool/TableHistograms.java
@@ -57,11 +57,15 @@ public class TableHistograms extends NodeToolCmd
         {
             checkArgument(false, "tablehistograms requires keyspace and table name arguments");
         }
-
-        // calculate percentile of row size and column count
-        long[] estimatedPartitionSize = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedPartitionSizeHistogram");
-        long[] estimatedColumnCount = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedColumnCountHistogram");
-
+        long[] estimatedPartitionSize;
+        long[] estimatedColumnCount;
+        try {
+            // calculate percentile of row size and column count
+            estimatedPartitionSize = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedPartitionSizeHistogram");
+            estimatedColumnCount = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedColumnCountHistogram");
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Unknown table: " + table + " or keyspace: " + keyspace);
+        }
         // build arrays to store percentile values
         double[] estimatedRowSizePercentiles = new double[7];
         double[] estimatedColumnCountPercentiles = new double[7];


slivne:
@elcallio please review

bhalevy:
@elcallio / @penberg: review ping

bhalevy:
Note: this should fix scylladb/scylla-jmx#55

src/java/org/apache/cassandra/tools/nodetool/TableHistograms.java
```diff
@@ -25,6 +25,8 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import javax.management.InstanceNotFoundException;
```
penberg: You are importing `InstanceNotFoundException` here, but...

src/java/org/apache/cassandra/tools/nodetool/TableHistograms.java
```diff
@@ -57,11 +59,16 @@ else if (args.size() == 1)
...
-        long[] estimatedPartitionSize = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedPartitionSizeHistogram");
-        long[] estimatedColumnCount = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedColumnCountHistogram");
-
+        long[] estimatedPartitionSize;
+        long[] estimatedColumnCount;
+        try {
+            // calculate percentile of row size and column count
+            estimatedPartitionSize = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedPartitionSizeHistogram");
+            estimatedColumnCount = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedColumnCountHistogram");
+        } catch (Exception e) {
```
penberg: ...using `Exception` here. Please either remove the import or fix this catch statement.

amnonh:
@penberg thanks for the comment, I've forced push a new version

src/java/org/apache/cassandra/tools/nodetool/TableHistograms.java
```diff
@@ -57,11 +57,16 @@ else if (args.size() == 1)
...
+            System.out.println("Keyspace/table not found");
```
penberg: The proposed fix for CASSANDRA-14410 has nicer error reporting, but since it's not upstream, let's worry about it later.

However, looking at their fix, you should throw `IllegalArgumentException` here and let top-level exception handler report the error. Also, please include keyspace and table name in the error message even if we don't know which one is missing.

amnonh: I've forced push a change with both throwing an exception and taking the message from the future upstream.

src/java/org/apache/cassandra/tools/nodetool/TableHistograms.java
```diff
@@ -57,11 +57,15 @@ else if (args.size() == 1)
...
-        long[] estimatedColumnCount = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedColumnCountHistogram");
-
+        long[] estimatedPartitionSize;
+        long[] estimatedColumnCount;
+        try {
+            // calculate percentile of row size and column count
+            estimatedPartitionSize = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedPartitionSizeHistogram");
+            estimatedColumnCount = (long[]) probe.getColumnFamilyMetric(keyspace, table, "EstimatedColumnCountHistogram");
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Unknown table: " + table + " in keyspace: " + keyspace);
```
penberg: Is this error message correct, though? AFAICT, we can fail if either the table or keyspace is missing, right? So the error message should be "Unknown table <table name> or keyspace <keyspace name>"?

amnonh: you right, it's probably better

PR #96, opened 2019-06-18
reorganize jar files so that offline installation works

glommer:
We put the .jar files for cassandra and scylla into the top level of the
.tar directory, and the dependencies inside lib/ The classpath is able
to find the latter, but not ther former, and as a consequence we cannot
execute the commands in bin/ and tools/bin -> they will fail on missing
jars.

If we move all jars - including the ones we build - to lib/, then
everything works in a nice, self-contained manner.

Once we do that, we can also simplify the installation for .deb and .rpm
packages, since they already move the entire contents of lib/ into the
filesystem.

Tested by installing the resulting rpm and deb, as well as using
nodetool, sstableloader and cassandra-stress directly from the resulting
.tar package.


penberg:
Looks ok to me, but can you @syuu1228 also have a look at it, please?

glommer:
ping @penberg and @syuu1228

syuu1228:
@glommer Looks fine for me.

glommer:
@avikivity it seems like @penberg left on vacations.

Who should merge this ?

PR #97, opened 2019-06-18
install.sh: install fakeroot

glommer:
fakeroot is needed to build debian packages on fedora, and it is not
installed by default. It is probably installed by scylla, so in a system
where all packages are being built this will likely be fine, but trying
to build just the tools will lead to failure.


penberg:
Btw, in scylla.git, we install the dependency in `dist/debian/build_deb.sh`, but I do think `install-dependencies.sh` makes more sense.

PR #103, opened 2019-07-12
make sure cqlsh works from the .tar.gz installation

### LINKED ISSUE #102: cqlsh does not work from the .tar.gz installation
Some of the users of our relocatable packages are rootless users and they do not consume the tools from an .rpm or .deb file.

cqlsh is currently broken in those setups:

```
Traceback (most recent call last):
  File "/tmp/tools/bin/cqlsh.py", line 169, in <module>
    from cqlshlib import cql3handling, cqlhandling, pylexotron, sslhandling, cqlshhandling
ImportError: No module named cqlshlib
```

The problem is that the the original cassandra code explicitly expects cqlshlib to be in the path `pylib/cqlsh` (or in the global system path):

```
# cqlsh should run correctly when run out of a Cassandra source tree,
# out of an unpacked Cassandra tarball, and after a proper package install.
cqlshlibdir = os.path.join(CASSANDRA_PATH, 'pylib')
if os.path.isdir(cqlshlibdir):
    sys.path.insert(0, cqlshlibdir)
```

Our current relocatable package runs the `setup.py` script to install cqlshlib inside `pylib/build`, and  later moves the contents of `pylib/build` to a system location. Only `pylib/build` is present in the final relocatable package.

This means that if a user is running cqlsh extracting from the relocatable package directly, cqlshlib will not be in the expected location and cqlsh will fail to execute.

glommer:
Some of the users of our relocatable packages are rootless users and they do
not consume the tools from an .rpm or .deb file.
cqlsh is currently broken in those setups:

Traceback (most recent call last):
  File "/tmp/tools/bin/cqlsh.py", line 169, in <module>
    from cqlshlib import cql3handling, cqlhandling, pylexotron, sslhandling, cqlshhandling
ImportError: No module named cqlshlib

The problem is that the the original cassandra code explicitly expects cqlshlib
to be in the path pylib/cqlsh (or in the global system path):

cqlshlibdir = os.path.join(CASSANDRA_PATH, 'pylib')
if os.path.isdir(cqlshlibdir):
    sys.path.insert(0, cqlshlibdir)

Our current relocatable package runs the setup.py script to install cqlshlib
inside pylib/build, and  later moves the contents of pylib/build to a
system location. Only pylib/build is present in the final relocatable
package.

This means that if a user is running cqlsh extracting from the relocatable
package directly, cqlshlib will not be in the expected location and cqlsh will
fail to execute.

The solution that this patch pursues is to make sure that the original cqlshlib
is the one included in the relocatable package, and the install phase, that deals
with the global paths, is moved to the .rpm or .deb packaging, where that actually
matters.

Fixes #102


glommer:
ping @penberg @syuu1228

glommer:
@penberg @syuu1228 hello guys, how are you doing today ?

penberg:
@glommer I changed merge target to `next`.

dist/debian/rules.mustache
```diff
@@ -20,6 +20,9 @@ build:
 install:
 	dh_testdir
 	dh_testroot
+	mkdir -p $(CURDIR)pylib/usr
+	cd $(CURDIR)/pylib && python setup.py install --no-compile --root $(CURDIR)/pylib/usr
+	cd $(CURDIR)
```
penberg: Is this last `cd` actually needed?

glommer: My understanding is that since we `cd` to a child directory in the last command, we need to cd back to the directory we were.

There are other commands executed after that and they may end up operating in the wrong directory

Unless this is executing each line in independent shells

PR #106, opened 2019-07-24
Implement sstable_info nodetool command

elcallio:
Requires seastar patch:
 json: Make date formatter use RFC8601/RFC3339 format

Requires scylla patch set "Implement `sstable_info` API command (info on sstables)"

Requires scylla-jmx patch set "sstable_info"

Adds nodetool "sstableinfo" command which simply forwards via JMX to scylla, and presents and formats the resulting tabular data. 

Note this patch set does not add domain object or change the MBean interfaces in tools-java. Instead we use loosly coupled call + composite unpacking.


slivne:
waiting for https://github.com/scylladb/scylla-jmx/pull/77

PR #108, opened 2019-08-15
cassandra.in.sh: copy only config file to conf

### LINKED ISSUE #94: Selectively copy configuration/property files for cassandra tools
In https://github.com/scylladb/scylla-tools-java/blob/master/dist/common/cassandra.in.sh#L14 we copy from $SCYLLA_CONF everything 

We have seen cases in which some non expected files are located in this directory and that causes issues - lets make the copy more selective and copy what is needed

amnonh:
This patch changes how configuration files are copied to the config
directory.

To prevent unneeded large file copy, only yaml files are copied.

Fixes #94

### DIFF
--- a/dist/common/cassandra.in.sh
+++ b/dist/common/cassandra.in.sh
@@ -2,6 +2,13 @@
 SCYLLA_HOME=/var/lib/scylla
 SCYLLA_CONF=/etc/scylla
 
+function cp_conf_dir {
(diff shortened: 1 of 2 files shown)


amnonh:
@syuu1228 can you take a look?

slivne:
@amnonh ping

amnonh:
@slivne I've forced pushed an updated version

elcallio:
LGTM

PR #114, opened 2019-09-07
Fix cqlsh on debian installations

### LINKED ISSUE #112: cross-build causes debian package to use unrecoginized site-packages, losing cqlsh
debian overrides the install location to dist-packages, but fedora doesn't know about that, so a cross-build ends up installing to site-packages, which debian doesn't pick up.

As a result, cqlsh doesn't work.

### LINKED ISSUE #113: cqlshlib files installed to incorrect location on debian
The debian package installs files to /usr/usr/python2.7 instead of /usr/python2.7.

avikivity:
This series fixes two bugs related to cqlsh installation on debian systems. First, debian doesn't use the standard `site-packages` directory; instead it uses `dist-packages`. Debian python distutils is patched appropriately and installs to the correct directory, but if we build the debian package on Fedora (as we do) the patch is not there, and the files are installed to the wrong directory and not picked up by python.

Second, 87dab77c1e797 ("make sure cqlsh works from the .tar.gz installation") had a mistake in its scylla-tools.install file which caused files which were meant for /usr into /usr/usr. As might be expected, python didn't pick up the files from there either.

Fixes #112 
Fixes #113 

Ref scylladb/scylla#4738

### DIFF
--- a/dist/debian/debian/scylla-tools.install
+++ b/dist/debian/debian/scylla-tools.install
@@ -12,4 +12,4 @@ tools/bin/sstablemetadata usr/bin
 tools/bin/sstablerepairedset usr/bin
 tools/bin/sstablesplit usr/bin
 dist/common/nodetool-completion etc/bash_completion.d/nodetool
-pylib/usr/* usr
+pylib/root/* .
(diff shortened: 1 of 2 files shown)


bhalevy:
who should review this?

slivne:
@penberg / @syuu1228  can you please review this is blocking 3.1

PR #115, opened 2019-09-20
None of the nodetool commands are working on master AMI

tarzanek:
fixes #5067

Port for nodetool command was not filled in.

/etc/scylla/cassandra/cassandra-env.sh fills in this port, so
source files need to copy it to temporary directory

### DIFF
--- a/dist/common/cassandra.in.sh
+++ b/dist/common/cassandra.in.sh
@@ -7,6 +7,7 @@ function cp_conf_dir {
     cp -a "$1"/*.xml "$2"  2>/dev/null || true
     cp -a "$1"/*.options "$2"  2>/dev/null || true
     cp -a "$1"/*.properties "$2"  2>/dev/null || true
+    cp -a "$1"/cassandra-env.sh "$2"  2>/dev/null || true
 }
 
 # Scylla adaption. Some one will still have to find us SCYLLA_HOME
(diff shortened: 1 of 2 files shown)


penberg:
@tarzanek Any idea what commit broke this?

tarzanek:
yes, it was broken by https://github.com/scylladb/scylla-tools-java/commit/c9630cf5bf1bbd08f426c7a34b4c5975b471209a
let me fix the pointer to fixed bug

tarzanek:
this PR fixes https://github.com/scylladb/scylla/issues/5067

dist/common/cassandra.in.sh
```diff
@@ -7,6 +7,7 @@ function cp_conf_dir {
     cp -a "$1"/*.xml "$2"  2>/dev/null || true
     cp -a "$1"/*.options "$2"  2>/dev/null || true
     cp -a "$1"/*.properties "$2"  2>/dev/null || true
+    cp -a "$1"/cassandra-env.sh "$2"  2>/dev/null || true
```
bhalevy: If it's mandatory why mask the error?

bhalevy: @tarzanek

tarzanek: hmm, that is true, it could be without error masking, though copying .yaml is "masked" too, can we open a bug for @amnonh to revisit this? I just continued with extending his patch ( https://github.com/scylladb/scylla-tools-java/pull/108 ), we can revisit his patch

PR #120, opened 2019-11-14
filter_cassandra_attributes.py: migrate from unsafe yaml.load()

ultrabug:
Read https://msg.pyyaml.org/load for full details (CVE-2017-18342)

Fix it by use new safe interface - yaml.safe_load()


avikivity:
Queued as 35906df2a0d1b94ab0df095dffd7f8c4ecf33e3a

PR #121, opened 2019-11-15
Nodetool over REST

tarzanek:


tarzanek:
@elcallio @amnonh @haaawk hi guys, if you will have a moment, please review
diff to pull request 100 (original one) - added data holder jmxbeans to get cfstats working without jmx (thnx Calle), added all jars as dependencies to lib (thnx Piotr)

this is the first step, basically after this moving all the rest of RESTNodeProbe to REST is just a matter of moving the rest calls from scylla-jmx to the probe (I don't expect such troubles as I had with cfstats dtests)

also @bhalevy  - I've only run dtest in byo , is there some better test suites to run for nodetool compliance? (so it behaves and reports same values like the old one?)

tarzanek:
reruning dtest after adding all commands the were requested:
https://jenkins.scylladb.com/view/master/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/704/

amnonh:
I have a general concern, I think that copying the logic from scylla-jmx to the nodetool is the wrong way around it.

For the long term, I think that taking the objects from scylla-jmx and plug them inside node_probe is the better approach.

This way is more robust to changes both in the API and in nodetool.

tarzanek:
@amnonh  hmm, I agree, but that would mean I will need to refactor scylla-jmx even more and break it again (same way how I extracted api client)
would that be feasible?

tarzanek:
@amnonh also I will have to make weird changes to scylla-jmx, since lots of mxbeans have private access + lack simple constructors (and I will need to add compatibility constructors)
so the impact will be that the scylla-jmx mxbeans will be more open (and possible unsafe, but this can be a lesser concern than duplicating of logic/code)

tarzanek:
@amnonh @elcallio 
so I have a new plan based on the talks we had and my experiments
(the move of mbeans to be used directly by nodetool didn't really work, standalone fake mbean server inside nodetool too)

1. extract needed REST stuff in scylla-jmx into helper objects (so we avoid duplication of REST calls implementation)
2. both scylla-jmx and scylla-tools-java will leverage the

build.xml
```diff
@@ -641,6 +660,18 @@
...
+        <dependency groupId="javax.ws.rs" artifactId="javax.ws.rs-api" version="2.0.1" />
+        <dependency groupId="javax.ws.rs" artifactId="jsr311-api" version="1.1.1" />
+        <dependency groupId="org.glassfish.jersey.core" artifactId="jersey-client" version="2.22.1" />
+        <dependency groupId="org.glassfish" artifactId="javax.json" version="1.0.4" />
+        <dependency groupId="javax.activation" artifactId="activation" version="1.1" />
+        <dependency groupId="com.google.collections" artifactId="google-collections" version="1.0" />
+        <dependency groupId="javax.activation" artifactId="activation" version="1.1" />
+        <dependency groupId="javax.annotation" artifactId="javax.annotation-api" version="1.2" />
+        <dependency groupId="org.glassfish.jersey.bundles.repackaged" artifactId="jersey-guava" version="2.22.1" />
+
```
elcallio: Good. Maven deps. 
But why are some (also maven avail) files binaries below?

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "scylla-jmx"]
+	path = scylla-jmx
+	url = git@github.com:scylladb/scylla-jmx.git
+	branch = master
```
elcallio: Can't we avoid submodules + build?
The cassandra build.xml has support for fetching maven dependencies. We should use this with a local maven repo (easy to set up in jenkins).

src/java/org/apache/cassandra/tools/NodeTool.java
```diff
@@ -147,9 +157,10 @@ public static void main(String... args)
                 // Remove until proven otherwise: RefreshSizeEstimates.class
                 // Remove until proven otherwise: RelocateSSTables.class,
                 ViewBuildStatus.class,
-                
                 SSTableInfo.class
-        );
+        ));
+
+        if (REST) { commands.add(RESTInfo.class); } else { commands.add(Info.class); }
```
elcallio: This is not accepable formatting. 
if (...) {
 ...
} else {
 ...
}
Always braces.

build.xml
tarzanek: ant and local run, it still expects libraries to be copied to lib properly (so above code is only published to exported maven poms, not really used locally :-( )

elcallio: Should be easy enough to use the maven fetch to add stuff to class path... Or simply add _all_ maven deps to compile class path?

.gitmodules
tarzanek: I had this talk with Pekka and all repos need to build (ideally) independently + we are not publishing maven artifacts from scylla-jmx and there is no plan for our own maven server
so I guess this could be solved by pushing all consumable artifacts from scylla-jmx to maven repos and having a proper scylla-jmx release cycle ... but since this doesn't make sense and is extra work we ended up deciding that submodule will be the easiest thing to do

src/java/org/apache/cassandra/tools/NodeTool.java
tarzanek: will fix

.gitmodules
elcallio: You can easily have automatic maven repo generation in jenkins via the maven-repo plugin. Would literally be 5 minutes work to set up for our maven projects.

tarzanek:
this PR is kinda waiting for something else - I am duplicating code,
which Amnon didn't like and he is right, ideally if we could take code from scylla-jmx and reuse it
Where I got stuck is - when I move the code away from the namespace, the jmx server in scylla jmx has some issues starting.
(I had to move the code, to be able to consume it + so it won't conflict with code and classes in this repo with the classloader)

slivne:
the is another issue with this - we need to support security - I am not sure we have this for the rest-api

mykaul:
Should be closed, native nodetool is almost ready.

PR #122, opened 2019-11-19
Switch to shard aware driver

haaawk:
Scylla has it's own shard aware driver and it can be
used to optimise connections to the cluster and
minimize cross-shard operations.

Signed-off-by: Piotr Jastrzebski <haaawk@gmail.com>


haaawk:
@avikivity @roydahan @penberg Please have a look.

PR #124, opened 2019-11-27
Make cqlshlib compatible with Python 3

fruch:
This fixes some issues in cqlshlib, which make it incompatible with 
Python 3.

We eventually need to make the whole cqlsh work with Python 3 similar to 
what upstream Apache Cassandra is doing:

https://issues.apache.org/jira/browse/CASSANDRA-10190

However, the main motivation for this change is to make sure we can port 
users of cqlshlib (such as scylla-dtest) to Python 3 now.

Tested on both Python 2 and Python 3.

### DIFF
--- a/pylib/cqlshlib/formatting.py
+++ b/pylib/cqlshlib/formatting.py
@@ -485,7 +485,7 @@ def format_value_text(val, encoding, colormap, quote=False, **_):
     if quote:
         bval = "'%s'" % bval
 
-    return bval if colormap is NO_COLOR_MAP else color_text(bval, colormap, wcwidth.wcswidth(bval.decode(encoding)))
+    return bval if colormap is NO_COLOR_MAP else color_text(bval.decode(encoding), colormap, wcwidth.wcswidth(bval.decode(encoding)))
 
 
 # name alias
--- a/pylib/cqlshlib/util.py
+++ b/pylib/cqlshlib/util.py
@@ -15,14 +15,22 @@
 # limitations under the License.
 
 
+from __future__ import print_function
 import cProfile
 import codecs
 import pstats
 
-from itertools import izip
+try:
+    from itertools import izip
+except ImportError:
+    izip = zip
+    
 from datetime import timedelta, tzinfo
-from StringIO import StringIO
-
+try:
+    from StringIO import StringIO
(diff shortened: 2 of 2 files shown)


fruch:
@elcallio @amnonh @haaawk, would love to get this one review
(a few more tests that would pass, when i'm moving dtest to python3 :smile: )

pylib/cqlshlib/formatting.py
```diff
@@ -485,7 +485,7 @@ def format_value_text(val, encoding, colormap, quote=False, **_):
     if quote:
         bval = "'%s'" % bval
 
-    return bval if colormap is NO_COLOR_MAP else color_text(bval, colormap, wcwidth.wcswidth(bval.decode(encoding)))
+    return bval if colormap is NO_COLOR_MAP else color_text(bval.decode(encoding), colormap, wcwidth.wcswidth(bval.decode(encoding)))
```
avikivity: Will this work correctly in python2? Now it will return a unicode object, no?

fruch: I've run cqlsh tests we have in dtest with this code and they passed. I can give cqlsh a go and check some Unicode behavior.

fruch: Looks like Cassandra team is working on python3 support for few years now

https://issues.apache.org/jira/plugins/servlet/mobile#issue/CASSANDRA-10190

fruch: I've test cqlsh manually with some unicode data, and it seems working o.k.
I think it was a bug that working with python3 just surfaced up.

in cassandra work for porting cqlsh to python, seem like they remove the encoding/decoding completely from this function

fruch: @avikivity if you want to be on the safe side here, I can put this change under if for python3 only.
FYI, this PR is not porting all of cqlsh to work python3, it's just for the test to be able to use the formatters.

bentsi:
@avikivity we tested the changes in Python 2 and Python 3. 
We need this PR to move dtest to Python 3
cc: @roydahan

fruch:
Title might be a bit misleading, since only specific part now support python3... Not cqlsh as a whole.

PR #130, opened 2019-12-18
driver: Use shard-aware driver by default

jul-stas:
Classpath contained Datastax driver and used it by default, meanwhile maven downloaded shard-aware driver but didn't use it.
It might have been misleading, so this commit replaces
`./lib/cassandra-driver-core-3.5.1-shaded.jar`
  with
`./lib/scylla-driver-core-3.7.1-scylla-2-shaded.jar`
(Note: both files are similar, e.g. my git client incorrectly considers this change as a "rename")

Refs scylladb/scylla#5406


haaawk:
@penberg @avikivity Could you merge this, please?

PR #132, opened 2020-01-05
Fix wcwidth.py to be python3 importable

fruch:
one more python3 related fix to make dtest some tests working


bhalevy:
please update target branch to `next`

bhalevy:
Also, add-
Fixes scylladb/scylla-dtest#1224 so that github can do the bookkeeping for us.

PR #134, opened 2020-01-08
dist/common/cassandra.in.sh: add cassandra_storagedir

syuu1228:
On our package, we specify empty parameter on "-Dcassandra.storagedir"  for
Java based programs, it because dist/common/cassandra.in.sh does not
initialized it.
Seems like tools/bin/cassandra.in.sh does initialize it, we can do same in
dist/common/cassandra.in.sh too.

(dist/common/cassandra.in.sh is for package, tools/bin/cassandra.in.sh is for
built from source code)

### DIFF
--- a/dist/common/cassandra.in.sh
+++ b/dist/common/cassandra.in.sh
@@ -2,6 +2,9 @@
 SCYLLA_HOME=/var/lib/scylla
 SCYLLA_CONF=/etc/scylla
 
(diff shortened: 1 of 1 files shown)


syuu1228:
cc: @glommer

penberg:
@syuu1228 What issue does this fix?

PR #135, opened 2020-01-09
stress: Remove "COMPACT STORAGE" from default c-s schema

roydahan:
COMPACT STORAGE is not used by default when creating a table in Scylla.
It's also not recommended anymore and cassandra supports an alter command to remove it but scylla don't support it yet.
The usage of compact storage have some limitations like adding and dropping columns.


roydahan:
@penberg / @slivne, can you please review?

I think it's about time we will do it...
After 3 years struggling with this default compact storage, it'll be better if we test scylla without it.
Currently, the test need to take care of it and it's a mess.

PR #141, opened 2020-01-27
Make C* compile and run on Java 11 and Java 8 (again)

elcallio:
patch by Robert Stupp; reviewed by Jason Brown for CASSANDRA-9608

reworked for scylla and 3.11 code base.

2020-01-27: fixed sstableloader script


penberg:
Packaging fails as follows:

```
15:17:54  install: cannot stat 'conf/jvm.options': No such file or directory
15:17:54  error: Bad exit status from /var/tmp/rpm-tmp.HEAdap (%install)
```

The `jvm.options` file seems to be renamed, but I am unsure what the correct one to take is, so I'm leaving this to you @elcallio. Looks like Debian package is also probably affected.

elcallio:
`jvm.options` was turned into:

- `jvm11-clients.options`
- `jvm8-clients.options`  
- `jvm-clients.options`
- `jvm11-server.options`
- `jvm8-server.options`
- `jvm-server.options`

where the 'client' ones should be all we care about. The reason I imagine is because some previously core packages in the jdk were turned into optional packages. The `11` ones enable these. 

Packaging should be updated to copy all these. The launch scripts already pick the appropriate one.

elcallio:
If you want me to modify packaging, you will first have to explain to me where it is. And maybe how it works...

penberg:
You can build the packages (and reproduce the problem) with the same instructions documented in `scylla.git`:

https://github.com/scylladb/scylla/blob/master/docs/building-packages.md

For this specific error, the old filename is in `install.sh` -- didn't check Debian packaging.

I can also modify the packaging if you so prefer if it's a small matter of including _all_ the properties file in the package. Let me know which way you prefer, @elcallio.

elcallio:
Oh, it is in the local install.sh file? I thought it was somewhere else out in cyberspace... :-) I've updated the script to use a wildcard instead. Verified with local install run.

penberg:
@elcallio I still see `jvm.options` in these files:

```
dist/debian/debian/scylla-tools-core.install:etc/scylla/cassandra/jvm.options
dist/redhat/scylla-tools.spec.mustache:%{_sysconfdir}/scylla/cassandra/jvm.options
```

Did you verify that the files are in the RPM package with `rpm -qlp <rpm filename>`?

elcallio:
*sigh* - did not notice all the duplicated patterns, no. And I cannot read my own `grep` results apparently...
Should be ok now. rpm  and deb builds verified. Sorry about the confusion.

penberg:
No worries, computers are hard for me too. Thanks, I'll spin it through bring-your-own to see what happens.

penberg:
This pull request broke two dtests, so I dequeued.

penberg:
@elcallio I sent you the dtest failures, can you please have a look and open a new pull request?

elcallio:
It is not an issue in the PR. It is CCM that needs fixing. I'll send a ccm patch.

PR #145, opened 2020-02-27
install.sh: add dependency check for manual install

syuu1228:
To install scylla-tools using install.sh easily, we need:
 - run dependency check before install

But we don't want to run them when we build .rpm/.deb package,
we also need to add --packaging option to skip them.

See scylladb/scylla#5830

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -31,6 +31,7 @@ Options:
   --prefix /prefix         directory prefix (default /usr)
   --etcdir /etc            specify etc directory path (default /etc)
   --nonroot                install Scylla without required root priviledge
+  --packaging               use install.sh for packaging
   --help                   this helpful message
 EOF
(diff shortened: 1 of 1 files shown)


syuu1228:
@avikivity this is scylla-tools part of merged PR https://github.com/scylladb/scylla/pull/5873 pending for few months, please review

avikivity:
Dequeued, has dependency on python2:

```
23:26:24  + ./install.sh --root /tmp/jenkins-e6b4f726/workspace/scylla-master/next/scylla-tools-java/build/redhat/BUILDROOT/scylla-tools-666.development-20200608.75e71d1b35.x86_64
23:26:24  Please install following package before running install.sh:
23:26:24    python2
23:26:24    python2-pyyaml
```

PR #151, opened 2020-04-13
cql_parsing: remove old quoted string rules

### LINKED ISSUE #150: via cqlsh not possible to escape special characters such as ' when used as FILTER for UPDATE
This is a copy of scylladb/scylla#6165

Summary of the problem is that in certain situations a pg style string that
contains single quotation mark is miss parsed in cqlsh and causes it to wait for more
input, the example in the issue was:
`UPDATE pi.upui SET createddate='1974-10-17 00:00:00.00+0000' WHERE upuishort=$$uyuyueyy'euyruwey$$;`

For more details refer to  scylladb/scylla#6165

eliransin:
This commit fixes an aggregated mistake in merges
from upstream. The string parsing rules for pg style
string have been mistakenly were added instead of being
replaced. Since the lexer works in a method of first mach
wins, the old parsing rules which were buggy persisted.
This caused to a mistake in parsing pg style strings that
contained single quotation mark.
This commit simply removes them so the new correct rules
will be used to parse the string.

Fixes #150

### DIFF
--- a/pylib/cqlshlib/cql3handling.py
+++ b/pylib/cqlshlib/cql3handling.py
@@ -143,8 +143,6 @@ JUNK ::= /([ \t\r\f\v]+|(--|[/][/])[^\n\r]*([\n\r]|$)|[/][*].*?[*][/])/ ;
 
 <stringLiteral> ::= <quotedStringLiteral>
                   | <pgStringLiteral> ;
-<quotedStringLiteral> ::= /'([^']|')*'/ ;
-<pgStringLiteral> ::= /(\$\$).*(\$\$)/ ;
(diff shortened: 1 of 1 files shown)


eliransin:
@penberg @elcallio I don't have permissions to add reviewers can you please review?
Thanks,

penberg:
Looks good to me, but let's ask @elcallio to review.

eliransin:
Thanks @elcallio :slightly_smiling_face: 
@penberg can you please merge?

penberg:
Queued to next as e813778bd1d308dd9fd8b0ec6a6f646c92cecd0a.

PR #152, opened 2020-04-14
Remove unused CDC logic that prints warnings

haaawk:
Cassandra and Scylla have totally different CDC implementations. Logic
removed by this patch is relevant only to Cassandra CDC which is never present
when working with Scylla. The code prints warnings that are confusing for
the users so it's better to remove it especially that the code is effectively
useless.

### DIFF
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -487,33 +487,6 @@ public class DatabaseDescriptor
             }
         }
 
-        if (conf.cdc_total_space_in_mb == 0)
-        {
-            int preferredSize = 4096;
-            int minSize = 0;
-            try
-            {
-                // use 1/8th of available space.  See discussion on #10013 and #10199 on the CL, taking half that for CDC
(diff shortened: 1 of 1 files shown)


haaawk:
@elcallio @penberg please review (unfortunately I can't add reviewers in this project)

haaawk:
> I'm ok with this, though removing the lines completely might cause merge issues down the road. But it is a risk I am willing to take. :-)

I was thinking about commenting it out but I think it will create merge conflicts too.

elcallio:
Yeah, I was more onto changing the config generation instead, to always set the cdc size. But that is probably more error prone, so this (remove) is good.

haaawk:
I agree. The alternative is to set some magic value to some magic option that's never used. It would be really confusing for a reader. Moreover, any changes to this Cassandra CDC logic in the future could break miserably due to the magic value we pick to set the option to. Much safer to just remove it.

eliransin:
@penberg can you please merge?

haaawk:
@penberg please

penberg:
Queued as 26a8d662a4d28575a0fb828145feacbcf0eb26d5, thanks @haaawk!

PR #153, opened 2020-04-22
Remove the confusing permission denied emitted at the start of nodetools output.

eliransin:
This commit handles the confusing permission denied when running nodetool  (and alse other tools) with a user that doesn't have write permissions to /opr/scylladb/share/cassandra

### DIFF
--- a/conf/cassandra-env.sh
+++ b/conf/cassandra-env.sh
@@ -85,6 +85,7 @@ calculate_heap_sizes()
         HEAP_NEWSIZE="${desired_yg_in_mb}M"
     fi
 }
+warn_gc_logs=false
 
 #GC log path has to be defined here because it needs to access CASSANDRA_HOME
 if [ $JAVA_VERSION -ge 11 ] ; then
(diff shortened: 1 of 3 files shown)


eliransin:
@avikivity can I please have permissions to triage and add reviewers on this repo?
@penberg @syuu1228 can you please review?

eliransin:
@syuu1228 ping

eliransin:
@penberg can you add takuya as a reviewer?

penberg:
@eliransin This should be reviewed by @elcallio, I think.

penberg:
@eliransin Can you please fix the pull request title to something else? I don't understand what it means... :-)

elcallio:
If we're running an install version of scylla, the `CASSANDRA_HOME` is gonna be something like `/opt/bla-bla/scylla-tools-java`, no? 
Would it not make more sense to put log files somewhere more sensible, like `/var/log` or something?
Do we even need any cassandra logs? Could we not always disable it given that the supported toolset should be limited to `nodetool` and `sstableloader`?

eliransin:
> Do we even need any cassandra logs? Could we not always disable it given that the supported toolset should be limited to `nodetool` and `sstableloader`?

@elcallio  I thought the same thing :slightly_smiling_face: , the only reason I didn't entirely loose this 
logic is in order to maintain compatibility with Cassandra. The only thing those logs collect is the 
JVM GC logs, I don't recall we ever had to deal with such problems.
Since you were the one that did the last merge from upstream and have the best guess if it can 
cause some merge problems down the road, do you think we should lose it entirely?

elcallio:
I think we should kill it, since it just creates junk logs that are of no consequence for neither developer or end user. But I am not really deploying things to customers, so someone else needs to sign off on that. 

We should maybe also try to separate behaviour of tool execution and server (though I hope `cassandra` script is not installed in our distro - but there is our testing and development, comparisons, also). 

Otoh, for this, the user (developer) can always add explicit log options to input env vars, so makes litte sense to leave gc log in scripts.

eliransin:
@vladzcloudius / @dyasny / @tarzanek (you really have to have a team in scylla so we can pin you all :slightly_smiling_face: ) please see https://github.com/scylladb/scylla-tools-java/pull/153#issuecomment-627818562 
This is about dropping creation of JVM GC logs by default in cassandra tools.
@elcallio and I are in favor but we will only drop it if you confirm that having this by default has no value for debugging tools failures.

conf/cassandra-env.sh
```diff
@@ -93,19 +94,37 @@ if [ $JAVA_VERSION -ge 11 ] ; then
...
+            warn_gc_logs=true
+        fi
     fi
 fi
 
+if $warn_gc_logs ; then
+    #put the warning to stderr so it will not interfere with any parsers of the command.
+    echo "WARNING: GC logs will not be collected" >&2
+    echo " Reason: ${CASSANDRA_HOME}/logs couldn't be found nor created." >&2
+    echo " This is probabely because you don't have sufficient permissions to crete directories in ${CASSANDRA_HOME}" >&2
```
tarzanek, line 122: "probably" , typo
"create" , typo

tarzanek:
other than typos, feel free to merge, gc logs for nodetool are useless so I agree with Calle

vladzcloudius:
> @vladzcloudius / @dyasny / @tarzanek (you really have to have a team in scylla so we can pin you all slightly_smiling_face ) please see [#153 (comment)](https://github.com/scylladb/scylla-tools-java/pull/153#issuecomment-627818562)
> This is about dropping creation of JVM GC logs by default in cassandra tools.
> @elcallio and I are in favor but we will only drop it if you confirm that having this by default has no value for debugging tools failures.

@eliransin I have no opinion on this hence I agree with the majority on this.

eliransin:
Closing this PR since we decided to go in a different direction:
See #163

PR #156, opened 2020-05-11
dist/redhat: drop dependency on pystache

syuu1228:
Same as https://github.com/scylladb/scylla/pull/6313, drop dependency on
pystache since it nolong present in Fedora 32.


penberg:
Queued to next as 4457d031ba3491ca8d4496157a77c176a469c6c3.

PR #163, opened 2020-05-14
Configuration: stop adding gc logging option to JVM by default

### LINKED ISSUE #147: nodetool: permission issues when running nodetool
This has been discovered by:
https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-centos7/34/
and a bunch of other artifact test on other distributions.
At some point, when running nodetool, cassandra-env.sh runs, this results in a failure on permission denied.
A recent commit : afd40c7 have added a creation of logs directory for the garbage collector and it 
tries to create it in $CASSANDRA_HOME those are the relevant lines:
https://github.com/scylladb/scylla-tools-java/blob/d3e5bbe208e1fd3dea7f4bdf82489c14b2f3e210/conf/cassandra-env.sh#L89-L108

In this particular failure, there is an attempt to create: /opt/scylladb/share/cassandra/logs which fails
on permission denied.

```
17:46:27  < t:2020-03-29 14:46:24,571 f:sct_events.py   l:791  c:sdcm.sct_events      p:INFO  > mkdir: cannot create directory ‘/opt/scylladb/share/cassandra/logs’: Permission denied
```
AFAICT this is a new behavior (the dir creation) but it might be necessary. So the options are:
1. (@syuu1228) the permissions on   `/opt/scylladb/share/cassandra` are not right.
2. (@elcallio) maybe there is something else that needs to be merged with this
3. There is a restriction on the user that should run nodetool (sudo / scylla user) - I couldn't find this in the docs.

/cc @roydahan

eliransin:
After considering fixing the "Permission Denied" print that is
triggered with every call to one of the java tools. It was
decided it is best to loose of not logging the JVM GC by default
at all.
The log has little to none value to us since we are not running
the actual cassandra server where it might have been of value
if it crashes.
If a developer/user still needs those logs he can add it as a
parameter to the JVM_OPTS environment variable.
For the discussion and previous proposed solution see:
https://github.com/scylladb/scylla-tools-java/pull/153

Fixes #147

### DIFF
--- a/conf/cassandra-env.sh
+++ b/conf/cassandra-env.sh
@@ -86,26 +86,6 @@ calculate_heap_sizes()
     fi
 }
 
-#GC log path has to be defined here because it needs to access CASSANDRA_HOME
-if [ $JAVA_VERSION -ge 11 ] ; then
-    # See description of https://bugs.openjdk.java.net/browse/JDK-8046148 for details about the syntax
-    # The following is the equivalent to -XX:+PrintGCDetails -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=10 -XX:GCLogFileSize=10M
-    echo "$JVM_OPTS" | grep -q "^-[X]log:gc"
-    if [ "$?" = "1" ] ; then # [X] to prevent ccm from replacing this line
-        # only add -Xlog:gc if it's not mentioned in jvm-server.options file
-        mkdir -p ${CASSANDRA_HOME}/logs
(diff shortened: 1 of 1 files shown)


elcallio:
lgtm

slivne:
@penberg please merge

slivne:
@eliransin I think this fixes #147 as well

slivne:
@penberg ping

penberg:
Queued as 80e8dab844262139bc0bd2b0ed8ce12cf43a26de.

slivne:
backported to 4.1 6ce3dd19252545c8ea74522a45f1e87225d1d53e

PR #165, opened 2020-05-20
NodeTool.java: enabled autocompaction cmds

fgelcer:
commands were implemented in Scylla, and
need to be enabled in nodetool to be used
by users.


fruch:
did you managed to run it ?

fgelcer:
> did you managed to run it ?

not yet

glommer:
You better hurry
tenor

fruch:
@glommer you were just looking for an excuse for using this gif, right ?

sitano:
I have compiled `nodetool` 80e8dab844262139bc0bd2b0ed8ce12cf43a26de with this patch and it really adds 2 new commands:

```
disableautocompaction   Disable autocompaction for the given keyspace and table
enableautocompaction    Enable autocompaction for the given keyspace and table
```

sitano:
Commands output for only existing `ks.loads`:

```
./bin/nodetool enableautocompaction      
./bin/nodetool enableautocompaction ka

nodetool: Keyspace [ka] does not exist.
See 'nodetool help' or 'nodetool help <command>'.

./bin/nodetool enableautocompaction ks
./bin/nodetool enableautocompaction ks ll

error: Scylla API server HTTP POST to URL '/storage_service/auto_compaction/ks' failed: std::_Nested_exception<no_such_column_family> (Can't find a column family ll in keyspace ks): std::out_of_range ()

-- StackTrace --
java.io.IOException: Scylla API server HTTP POST to URL '/storage_service/auto_compaction/ks' failed: std::_Nested_exception<no_such_column_family> (Can't find a column family ll in keyspace ks): std::out_of_range ()
	at org.apache.cassandra.service.StorageService.enableAutoCompaction(StorageService.java:1451)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
...
	at java.lang.Thread.run(Thread.java:748)

./bin/nodetool enableautocompaction ks loads
./bin/nodetool disableautocompaction         
./bin/nodetool disableautocompaction 
./bin/nodetool disableautocompaction ka

nodetool: Keyspace [ka] does not exist.
See 'nodetool help' or 'nodetool help <command>'.

./bin/nodetool disableautocompaction ks
./bin/nodetool disableautocompaction ks ll
./bin/nodetool disableautocompaction ks loads
./bin/nodetool disableautocompaction ks loads
```

Maybe you would like to have different error handling behavior for `enableautocompaction`. It may be fixed on different levels (JMX, NodeTool, Scylla).

fgelcer:
> however, those commands depend on the behavior of the `cfs.isAutoCompactionDisabled()` JMX result at `org.apache.cassandra.db.ColumnFamilyStore#isAutoCompactionDisabled`, so the `submitBackground()` that does not submit a MC if it thinks it is disabled. And as far current impl of cfs.isAutoCompactionDisabled at JMX module is invalid I expect it would not work properly. I think we must fix JMX cfs.isAutoCompactionDisabled first. (scylla API has the API point value inverted)

so @sitano , what are you planning to do regarding this? also, the patch to JMX (if you didn't do it yet) as you mentioned:
```
public boolean isAutoCompactionDisabled() {
log(" isAutoCompactionDisabled()");
return client.getBooleanValue("column_family/autocompaction/" + getCFName());
}
```

penberg:
The change looks OK, but it's not clear to me if this was sufficient or not. @sitano / @fgelcer, what's the status of this?

sitano:
@penberg we need to fix `isAutoCompactionDisabled` in JMX first. otherwise, nodetool compaction enable/disable would not work correctly. It can be verified with the trace logging in the nodetool and ScyllaDB REST API on compaction status.

Then, we need to decide on API behavior for the cases when the user passes invalid arguments like those of unexisting keyspaces or tables. At minimum `enableautocompaction` must not return a stacktrace if requested keyspace does not exist. I think the perfect fix would be to patch the JMX error handling of this end. Or we can also patch Scylla endpoint to return errors in a different way.

Then, from the usability standpoint, what is also concerns me, is that nodetool commands are silent about what they did if they did anything. However, this could be mimic C* nodetool version. It's worth checking.

I can look into (1) and (2) eg today. (3) need someone who has time. (3) is not blocker. (1) and (2) I would say worth fixing first.

sitano:
@penberg I rechecked code. The nodetool part just calls commands and directly passes the control to the MBean JMX. So I am sorry, it is fine, `isAutoCompactionDisabled` is not involved. PR is good to go. 

The only thing that we can improve is error handling.

sitano:
To verify it is working the following can be done:

JMX tracing
===

1) create `logging.properties`:

```
.level=INFO
java.util.logging.ConsoleHandler.level = FINEST
org.apache.cassandra.service.StorageService.level = FINEST
org.apache.cassandra.service.StorageService.handlers = java.util.logging.ConsoleHandler
sun.net.www.protocol.http.HttpURLConnection.level = FINE
sun.net.www.protocol.http.HttpURLConnection.handlers = java.util.logging.ConsoleHandler
```

2) execute `./scripts/scylla-jmx -Djava.util.logging.config.file=logging.properties`

ScyllaDB compaction status verification
===

```
$ curl http://127.0.0.1:10000/column_family/autocompaction/${keyspace}:${table}
```

returns `true` or `false` depending on the status

fgelcer:
@sitano , could you please approve this PR to have it merged by @penberg ?
thanks.

PR #166, opened 2020-05-25
config: [RFC] avoid checking options and filtering scylla.yaml

jul-stas:
The purpose of this change is to allow scylla.yaml being used
instead of cassandra.yaml. This is achieved simply by omitting
config checks in Java, but it appears to be enough.

Since scylla.yaml is now digestible to java tools, it was made the default.
Also, python scripts that do options filtering are no longer needed.

Fixes scylladb/scylla#2454


haaawk:
Why "Refs" and not "Fixes"?

haaawk:
jvm*.options are just files with JVM flags used to configure JVM that runs the tool. Those files shouldn't be in SCYLLA_CONF but in some place related to the tools. They control JVM for client so they are not related to Scylla server. Can we fix the script to look for those files in tools related dir?

dist/common/cassandra.in.sh
```diff
@@ -12,28 +12,17 @@ cp_conf_dir () {
...
-    fi
+    SCYLLA_CONFIG_FILE="scylla.yaml"
+    CASSANDRA_CONF=$SCYLLA_CONF
+fi
+if [ -f "$SCYLLA_CONF/cassandra.yaml" ]; then
+    SCYLLA_CONFIG_FILE="cassandra.yaml"
+    CASSANDRA_CONF=$SCYLLA_CONF
 fi
+echo "Using $SCYLLA_CONF/$SCYLLA_CONFIG_FILE as the config file"
+CONFIGURATION_FILE_OPT="-Dcassandra.config=file://$SCYLLA_CONF/$SCYLLA_CONFIG_FILE"
```
haaawk, line 25: Isn't this the reason why SCYLLA_CONF has to be a full path? Does "file://" work with relative paths?

dist/debian/debian/rules
```diff
@@ -17,9 +17,6 @@ override_dh_makeshlibs:
 override_dh_shlibdeps:
 
 override_dh_fixperms:
-	dh_fixperms
```
haaawk, line 20: @syuu1228 Does this look ok to you?

install.sh
```diff
@@ -113,14 +113,6 @@ install -m644 lib/licenses/* "$rprefix"/share/cassandra/lib/licenses
 install -m644 doc/cql3/CQL.css doc/cql3/CQL.html "$rprefix"/share/cassandra/doc/cql3
 ln -srf "$rprefix"/share/cassandra/lib/apache-cassandra-$scylla_version-$scylla_release.jar "$rprefix"/share/cassandra/lib/apache-cassandra.jar
 install -m644 dist/common/cassandra.in.sh "$rprefix"/share/cassandra/bin
-for i in tools/bin/{filter_cassandra_attributes.py,cassandra_attributes.py} bin/scylla-sstableloader; do
```
haaawk, line 116: Is scylla-sstableloader removed as well?

dist/common/cassandra.in.sh
jul-stas: Yes, that's the reason.
My experiments with relative paths in URI were unsuccessful. Apparently some implementations don't allow it: https://stackoverflow.com/questions/7857416/file-uri-scheme-and-relative-files

penberg:
Let's have @elcallio and @syuu1228 review this.

install.sh
jul-stas: Oops, fix'd.
But now that I look at it, this line might have been defunct:
```
install -m755 $i "$rprefix"/share/cassandra/bin
```
because `$i` is a basename. I'll test it on a VM.

jul-stas:
> Why "Refs" and not "Fixes"?

I wasn't sure if cross-repo "Fixes" work, but let's try it.

jul-stas:
> jvm*.options are just files with JVM flags used to configure JVM that runs the tool. Those files shouldn't be in SCYLLA_CONF but in some place related to the tools. They control JVM for client so they are not related to Scylla server. Can we fix the script to look for those files in tools related dir?

Sure, will do.

dist/common/cassandra.in.sh
elcallio:
```
CASSANDRA_CONF=`realpath $SCYLLA_CONF`
```

elcallio:
> 
> 
> jvm*.options are just files with JVM flags used to configure JVM that runs the tool. Those files shouldn't be in SCYLLA_CONF but in some place related to the tools. They control JVM for client so they are not related to Scylla server. Can we fix the script to look for those files in tools related dir?

jvm*.options should already be handled in packaging? Why does this patch need to change anything about them? They are doing unrelated things...

haaawk:
> Is this verified to really work? The filtering was not wholly about removing opts that don't exist in origin, but also dealing with some that have different syntax. For example, any server or client TLS options in a scylla.yaml will most likely cause the argument mapper (yaml beans populator) to at least partially barf.
> And, no, we don't use those things in the tools, but the argument parser might still throw.
> Exactly which permutations of scylla options was this verified with?
> 
> One more thing: While a worthy goal, an even better goal would be to remove the need to have cross-mapped options at all. Afaik, the only real reason it existed at any point it that sstabledump needed the node schema tables to operate. As of later sstable formats, this is no longer true. Are there actually any tooling that _need_ a "shared" config, or could we just get away with keeping a static, dummy conf for scylla-tools-java?

We're still supporting LA/KA sstables so sstabledump still needs schema sometimes.

dist/common/cassandra.in.sh
jul-stas: done

jul-stas:
@elcallio You were right about "extra" options. I got overly trigger-happy and pushed a version with removed property checker. Fixed - now options unknown to C* are quietly going through. Tested on default `scylla.yaml`.

Also, I just uncommented both groups of `{client,server}_encryption_options` in `scylla.yaml`, including empty ones. It worked, so... do you have any other troublesome options on your mind (i.e. ones worth checking)?

elcallio:
I am not sure which options we redefine syntax/values for so, no. A good guess is that anything mentioned in the original filter _might_ be troublesome. You should verify them all, with a reasonable permutation of values. Fun, right. :-)

jul-stas:
Couple of updates:

- I changed the logic a little: now `CASSANDRA_CONF` is untouched, therefore JVM configs are read properly - without warnings;
- I hardcoded `dist/common/cassandra.in.sh` to use `scylla.yaml` and to search for JVM opts in `/etc/scylla/cassandra`.
Do we ever have `cassandra.yaml` in `/etc/scylla`? I'm assuming that no.
- I tweaked the default location of `scylla.yaml`, assuming that `scylla` and `scylla-tools-java` live next to each other, so `scylla.yaml` is found and used by default;
- I restored `dh_fixperms` whatever that is;
- tested on `scylla.yaml` with randomly uncommented options, including all options - no issues found;

I'll do more testing on scyllas installed system-wide, but for now **please**, have a careful look. I'm new to installation/packaging topics and might have got something totally wrong.

dist/debian/debian/rules
jul-stas: I restored it, but opinion from @syuu1228 is still needed.

jul-stas:
@elcallio Yo Dawg, I heard you like scylla-tools-java, so I hereby ping.

install.sh
```diff
@@ -132,7 +125,7 @@ install -d -m755 "$rprefix"/share/cassandra/pylib
 install -d -m755 "$retc"/bash_completion.d
 install -m644 dist/common/nodetool-completion "$retc"/bash_completion.d
 cp -r pylib/cqlshlib "$rprefix"/share/cassandra/pylib
-for i in bin/{nodetool,sstableloader,cqlsh,cqlsh.py} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
+for i in bin/{nodetool,sstableloader,cqlsh,cqlsh.py,scylla-sstableloader} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
```
elcallio, line 128: Why two? We should never package the origin sstableloader, as it is completely useless for scylla....

haaawk: @slivne @syuu1228 Could you please comment on this?

haaawk:
@slivne is @syuu1228 the right person to review this as well?

install.sh
jul-stas: Well, I ended up with a for-loop with one element, so I merged it into adjacent one. I have no clue why it was there.

slivne:
lets ask @penberg

penberg:
Queued as 3eca0e35111f48629cf5b654dd5212199029f78e.

PR #167, opened 2020-05-25
reloc-pkg: move all files under project name directory

syuu1228:
*this is reopened version of https://github.com/scylladb/scylla-tools-java/pull/155, since it dequeued*

To make unified relocatable package easily, we may want to merge tarballs to single tarball like this:
zcat .tar.gz | gzip -c > scylla-unified.tar.xz
But it's not possible with current relocatable package format, since there are multiple files conflicts, install.sh, SCYLLA--FILE, dist/, README.md, etc..

To support this, we need to archive everything in the directory when building relocatable package.

See: scylladb/scylla#6315

### DIFF
--- a/dist/debian/build_deb.sh
+++ b/dist/debian/build_deb.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-PRODUCT=$(cat SCYLLA-PRODUCT-FILE)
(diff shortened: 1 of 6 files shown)


syuu1228:
@avikivity @penberg ping

penberg:
@syuu1228 Conflicts, please rebase.

syuu1228:
@penberg rebased, please review it again

PR #178, opened 2020-07-05
dist/debian: apply generated package version for .orig.tar.gz file

syuu1228:
Same as scylladb/scylla#6752,
we currently does not able to apply version number fixup for .orig.tar.gz file,
even we applied correct fixup on debian/changelog, becuase it just reading
SCYLLA-VERSION-FILE.
We should parse debian/{changelog,control} instead.

### DIFF
--- a/dist/debian/build_deb.sh
+++ b/dist/debian/build_deb.sh
@@ -91,12 +91,13 @@ if [ -z "$TARGET" ]; then
         exit 1
     fi
 fi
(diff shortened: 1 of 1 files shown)


syuu1228:
@avikivity As I commented on scylla-jmx, this is also required to fix scylla-4.2 build error

avikivity:
Backported to 4.2.

PR #180, opened 2020-07-07
dist: debian: fix detection of debuild

avikivity:
We detect if debuild is already installed, in order to avoid
re-installing it, which can fail due to flakey mirrors. However,
we use the wrong path to detect if it is installed, so we do in fact
fail on flakey Fedora mirrors.

Fix by using the correct path.


avikivity:
@penberg ping

penberg:
Queued as f57fbb77b0a3f8d240c9924f3fa4529f5b5c8122

PR #182, opened 2020-07-13
NodeProbe: get all histogram values in a single call

amnonh:
Table histagrams uses sampling to calculate the histogram.
There could be a potential problem when querying the JMX multiple times,
the values can potentiallay change so drastically that the histogram
values will not be monotonic.

This patch changes NodeProbe implementation, instead of querying each of
the quantile values from the JMX seperately, it will do a single request
to get the values and then will calculate the histogram itself using the
same aggregation class: BufferSamples as the jmx does.

Fixes scylladb/scylla#6792

Signed-off-by: Amnon Heiman <amnon@scylladb.com>

### DIFF
--- a/src/java/org/apache/cassandra/tools/NodeProbe.java
+++ b/src/java/org/apache/cassandra/tools/NodeProbe.java
@@ -17,6 +17,8 @@
  */
 package org.apache.cassandra.tools;
 
+import static java.lang.Math.floor;
+
 import java.io.IOException;
 import java.io.PrintStream;
 import java.lang.management.ManagementFactory;
@@ -30,6 +32,7 @@ import java.rmi.server.RMIClientSocketFactory;
 import java.rmi.server.RMISocketFactory;
 import java.util.AbstractMap;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -129,6 +132,56 @@
     private BatchlogManagerMBean bmProxy;
     private boolean failed;
 
+    private static class BufferSamples  {
```
avikivity, line 135: I see there is already a BufferSamples class elsewhere.

I don't understand where you get those samples from.

amnonh: You are right.

The nodetool uses the old jmx-histogram implementation (which I wanted to change, but I'll address that later). The idea is to hold `n` data samples, not a true histogram.

BufferedSamples is part of the java proxy (scylla-jmx) it takes those samples and calculates a histogram.

The issue this patch tries to solve, that the current implementation calls the jmx multiple times for a single histogram, so it could be that a call to fetch the 75% value will get a higher result than a call to the 99% value.

Instead, I've copied that functionality to NodeProbe (the JMX interface) so there will be a single call between the nodetool and the jmx, which will make the result consistent.

About switching to a normal histogram, I want to do it but there is a problem with derivation, nodetool does not look at the overall histogram, instead, it gives an approximation for the recent values.
To get this recent histogram we need state, we can do one of the following:
1. Add that to Scylla's own implementation by periodically takes a snapshot of the histogram and calculate the delta.
2. Add that to the nodetool, so a call will take a bit longer but it would ask for the results twice and calculate the difference. 3. We can query the Prometheus server that holds that history.

I prefer the second approach.

See also:
https://github.com/scylladb/scylla/issues/3468

slivne:
@elcallio can you please review

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -1475,13 +1528,15 @@ public long getStorageMetric(String metricName)
...
-                metric.getMax()};
+        BufferSamples bs = new BufferSamples(metric.values());
+
+        return new double[]{ bs.getValue(0.5),
+                bs.getValue(0.75),
+                bs.getValue(0.95),
+                bs.getValue(0.98),
+                bs.getValue(0.99),
+                bs.getMin(),
+                bs.getMax()};
```
elcallio, line 1539: The `metricPercentilesAsArray(JmxHistogramMBean)` before this method seem to have the same problem (`getSnapshot().values().get...`). Why did you not replace this with a calculation as well?

slivne:
@amnonh ping

src/java/org/apache/cassandra/tools/NodeProbe.java
amnonh: @elcallio Thanks for pointing it out, I've forced push a fix with this change.

avikivity:
I see the BufferSamples class is still duplicated.

amnonh:
On Wed, Aug 5, 2020 at 6:37 PM Avi Kivity <notifications@github.com> wrote:

> I see the BufferSamples class is still duplicated.
>
It's in different processes and codebase, one is in scylla-jmx the other is
in nodetool (nodeprobe).

avikivity:
> On Wed, Aug 5, 2020 at 6:37 PM Avi Kivity ***@***.***> wrote: I see the BufferSamples class is still duplicated.
> It's in different processes and codebase, one is in scylla-jmx the other is in nodetool (nodeprobe).
>

Yes :facepalm:

avikivity:
Please add an entry to NOTICE. See [guidlines](https://infra.apache.org/licensing-howto.html).

amnonh:
Just did, thanks

avikivity:
Please fold the two commits.

amnonh:
Sure, Folded

bhalevy:
@scylladb/scylla-tools-maint please consider backport to 4.2 and 4.1

PR #183, opened 2020-07-14
dist: redhat: reduce log spam from unpacking sources when building rpm

avikivity:
rpmbuild defaults to logging the name of every file it unpacks from
the archive. This is quite a lot for Java applications.

Make it quiet with the %setup -q flag.


penberg:
Queued as 113c7d993b038e13fd34a4a40b2aea7e324507f8

PR #188, opened 2020-07-21
nodetool: Use non-default `cassandra.config`

jul-stas:
My understanding is that it should work, since this test: https://github.com/scylladb/scylla/issues/6878 has this log message: 

`Using /etc/scylla/scylla.yaml as the config file`

so the variable `CONFIG_FILE_REALPATH` is set and thus `CONFIGURATION_FILE_OPT` is also set. But an experiment is strongly advised.

Fixes scylladb/scylla#6878

### DIFF
--- a/bin/nodetool
+++ b/bin/nodetool
@@ -98,6 +98,7 @@ fi
         -Dcassandra.storagedir="$cassandra_storagedir" \
         -Dlogback.configurationFile=logback-tools.xml \
         $JVM_ARGS \
+        $CONFIGURATION_FILE_OPT \
         org.apache.cassandra.tools.NodeTool -p $JMX_PORT $ARGS
 
 # vi:ai sw=4 ts=4 tw=0 et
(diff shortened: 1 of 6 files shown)


src/java/org/apache/cassandra/config/YamlConfigurationLoader.java
```diff
@@ -52,7 +52,7 @@
 {
     private static final Logger logger = LoggerFactory.getLogger(YamlConfigurationLoader.class);
 
-    private final static String DEFAULT_CONFIGURATION = "cassandra.yaml";
+    private final static String DEFAULT_CONFIGURATION = "file:///cassandra.yaml";
```
elcallio: `file:///cassandra,yaml` means file at `/cassandra.yaml.` I assume you did not mean file at root?
Actually, relative URL for `cassandra.yaml` is in fact `cassandra.yaml` . I,e, what it said before.

jul-stas: Oh, brainfart. Fixed.

elcallio:
lgtm

haaawk:
Is this change nodetool specific or should we fix other tools as well?

elcallio:
If you change all `bin/*` launchers, should you not also include the `tools/bin/*` ones as well?

jul-stas:
> Is this change nodetool specific or should we fix other tools as well?

@haaawk Sorry, I posted this as a quickfix and checked the other tools in the meantime.
I added the option where it seems needed.

@elcallio IIUC, we already did that in the previous patch.

penberg:
Queued as e62e76db299684e57c5931cf598701594207ad96

PR #192, opened 2020-08-02
dist: debian: support non-x86

avikivity:
The package is already arch independent, so remove the artifical
restriction to x86.


avikivity:
@penberg ping

avikivity:
@penberg ping

penberg:
Sorry, somehow missed this when I merged the JMX counterpart.

Queued as f2c7cf8d8d664e2c817df8ac4e8d63dd0b931e69.

PR #199, opened 2020-08-26
sstableloader: remove wrong check that breaks range tombstones

### LINKED ISSUE #198: sstableloader does not work with range tombstones
https://github.com/scylladb/scylla-tools-java/commit/14b99c050aab4dd380b2edb836c6c20350b839fe has introduced a check that breaks when sstable contains a range tombstone:

```
        private void setWhere(Slice.Bound start, Slice.Bound end) {
            assert where.isEmpty();
            
            ClusteringPrefix spfx = start.clustering();
            ClusteringPrefix epfx = end.clustering();
            
            List<ColumnDefinition> clusteringColumns = cfMetaData.clusteringColumns();
            for (int i = 0; i < clusteringColumns.size(); i++) {
                ColumnDefinition column = clusteringColumns.get(i);                
                
                Object sval = i < spfx.size() ? column.cellValueType().compose(spfx.get(i)) : null;
                Object eval = spfx != epfx ? (i < epfx.size() ? column.cellValueType().compose(epfx.get(i)) : null) : sval;
                
                if (sval == null && eval == null) {
                    // nothing. 
                } else if (sval != null && (sval == eval || sval.equals(eval))) {
                    assert start.isInclusive();
                    where.put(column, Pair.create(Comp.Equal, sval));                                                              
                } else {


BEGINING OF THE CHECK

>>>>>               if (column.isPrimaryKeyColumn()) {
>>>>>                   // cannot generate <> for pk columns
>>>>>                   throw new IllegalStateException("Cannot generate <> comparison for primary key columns");
>>>>>               }

END OF THE CHECK


                    if (sval != null) {
                        where.put(column, 
                                Pair.create( 
                                        start.isInclusive() ? Comp.GreaterEqual : Comp.Greater, sval));
                    }
                    if (eval != null) {
                        where.put(column, 
                            Pair.create( 
                            end.isInclusive() ? Comp.LessEqual : Comp.Less, eval));
                    }
                }
            } 
        }
```

The problem was reported by a user on our Slack.

The check does not make sense because every `column` in the loop is a clustering column so it's always part of primary key.
No range tombstone can pass this check.

@elcallio and I wasn't able to figure out why this check is here. The suspicion is that it's a result of a wrong rebase.

haaawk:
The check does not make sense because every column in the loop is a clustering
column so it's always part of primary key.
No range tombstone can pass this check.

Fixes #198

### DIFF
--- a/src/java/com/scylladb/tools/SSTableToCQL.java
+++ b/src/java/com/scylladb/tools/SSTableToCQL.java
@@ -399,10 +399,6 @@ public class SSTableToCQL {
                     assert start.isInclusive();
                     where.put(column, Pair.create(Comp.Equal, sval));                                                              
                 } else {
-                    if (column.isPrimaryKeyColumn()) {
-                        // cannot generate <> for pk columns
-                        throw new IllegalStateException("Cannot generate <> comparison for primary key colum " + column);
-                    }
                     if (sval != null) {
                         where.put(column, 
                                 Pair.create( 


penberg:
Merged as https://github.com/scylladb/scylla-tools-java/commit/2d49ded77b7dea47543b60e19d83f7e3362cfc4d

PR #201, opened 2020-09-09
sstableloader: Support range boundary markers

avelanarius:
Add support for range tombstone markers of `range_tombstone_boundary_marker` type. Previously only `range_tombstone_bound_marker` were properly handled.

### DIFF
--- a/src/java/com/scylladb/tools/SSTableToCQL.java
+++ b/src/java/com/scylladb/tools/SSTableToCQL.java
@@ -396,7 +396,7 @@ public class SSTableToCQL {
                         where.put(column, Pair.create(Comp.Equal, i < spfx.size() ? spfx.get(i) : epfx.get(i)));
                     }
                 } else if (sval != null && (sval == eval || sval.equals(eval))) {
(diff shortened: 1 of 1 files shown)


src/java/com/scylladb/tools/SSTableToCQL.java
```diff
@@ -819,6 +819,13 @@ private void process(UnfilteredRowIterator rows) {
         private void process(RangeTombstoneMarker tombstone) {
             ClusteringBound end = tombstone.closeBound(false);
 
+            if (end != null && tombstone.isBoundary()) {
+                ClusteringBound start = tombstone.openBound(false);
+                assert start != null;
+                deleteCqlRow(start, end, tombstone.openDeletionTime(false).markedForDeleteAt());
```
haaawk, line 825: This seems wrong. start and end don't belong to the same range tombstone.

avelanarius: As per our chat conversation, it is indeed incorrect. (comment for other reviewers)

PR #208, opened 2020-09-22
build: support passing product-version-release as a parameter

avikivity:
Instead of using the baked-in values from SCYLLA-VERSION-GEN,
allow passing an override. This will be used by the supermodule
to have an identical product-version-release (especially release,
which contains the git hash) across all packages.


avikivity:
Using wrong repo

PR #211, opened 2020-10-05
dist/debian/debian_files_gen.py: don't ignore permission error on shtil.rmtree()

syuu1228:
shutil.rmtree(ignore_errors=True) was for ignores error when directory not exist,
but it also ignores permission error, so we shouldn't use that.
Run os.path.exists() before shutil.rmtree() instead.


penberg:
@syuu1228 The pull request description does not seem to match the commit.

syuu1228:
@penberg Sorry, I copy & pasted wrong comment, fixed

PR #212, opened 2020-10-12
Relocatable Package: create product prefixed relocatable archive

eliransin:
The build system was hardcoded to produce a package that is
prefixed with scylla instead of the product name. This is not
in line with out CI system requirements and can be also a source
for confusion.
This commit make the packaging system generate a package of
the format: {product}-tools-package.tar.gz instead of
scylla-tools-package.tar.gz

### DIFF
--- a/reloc/build_rpm.sh
+++ b/reloc/build_rpm.sh
@@ -31,6 +31,7 @@ mkdir -p "$BUILDDIR"
 tar -C "$BUILDDIR" -xpf $RELOC_PKG scylla-tools/SCYLLA-RELEASE-FILE scylla-tools/SCYLLA-RELOCATABLE-FILE scylla-tools/SCYLLA-VERSION-FILE scylla-tools/SCYLLA-PRODUCT-FILE scylla-tools/dist/redhat
 cd "$BUILDDIR"/scylla-tools
 
+RELOC_PKG_BASENAME=$(basename "$RELOC_PKG")
 SCYLLA_VERSION=$(cat SCYLLA-VERSION-FILE)
 SCYLLA_RELEASE=$(cat SCYLLA-RELEASE-FILE)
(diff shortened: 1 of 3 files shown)


yaronkaikov (approved):

reloc/build_rpm.sh
```diff
@@ -44,6 +45,7 @@ parameters=(
     -D"version $SCYLLA_VERSION"
     -D"release $SCYLLA_RELEASE"
     -D"product $PRODUCT"
+    -D"reloc_pkg $RELOC_PKG_BASENAME"
 )
```
avikivity, line 50: What about .deb? Doesn't it need similar treatment?

eliransin: According to my tests no - I believe this is because of this line:
https://github.com/scylladb/scylla-tools-java/blob/f2e8666d7e850c8ed715cba3fc466e8ded21e844/reloc/build_deb.sh#L40
Which adjust the relocatable to whatever the debian packaging needs (I admit I am not en expert of this so I am also 
relying on empiric results and not pure knowledge).
Another thing to notice is that rpm is actually buggy and we just haven't encounter it before. The relocatable path is 
passed to build_rpm.sh  script but it is just being ignored by the spec, which means that until this patch it would have never
worked with relocatable packages named something other than scylla-tools-package.tar.gz :slightly_smiling_face:

eliransin:
@penberg ping (merge or comment :slightly_smiling_face: )
unless @avikivity or you are not convinced with my explanation why we don't need to change the build_deb procedure.

eliransin:
@scylladb/scylla-tools-maint can someone merge please.

eliransin:
@scylladb/scylla-tools-maint ping

nyh:
Committed - e97c1060476085d25f9709b94df8ab48eac03c39

PR #222, opened 2021-01-26
cassandra.in.sh: remove debug message

### LINKED ISSUE #213: cassandra.in.sh shouldn't print debug prints into stdout
like here:
https://github.com/scylladb/scylla-tools-java/blob/3eca0e35111f48629cf5b654dd5212199029f78e/tools/bin/cassandra.in.sh#L63

this shouldn't be part of the stdout, for example in tools like sstabledump, it would break some tools and automation with this outout

jul-stas:
Apparently, some automation tools rely on the output, so debug messages are out of place here.

Fixes #213


fruch (approved):
LGTM

Issue #218, opened 2021-01-25
sstableloader fails to restore backup files created by scylla manager


jtnagashima:
## Version:
sstableloader: 8080009794bb628380dbfed27e02cd2e653078b4

Scylla Cluster - Source Destination:
```
scylla-4.2.3-0.20210104.24346215c2.x86_64
scylla-conf-4.2.3-0.20210104.24346215c2.x86_64
scylla-debuginfo-4.2.3-0.20210104.24346215c2.x86_64
scylla-env-1.1-1.el7.noarch
scylla-jmx-4.2.3-20210105.e351b009da.noarch
scylla-kernel-conf-4.2.3-0.20210104.24346215c2.x86_64
scylla-libatomic73-7.3.1-1.2.el7.centos.x86_64
scylla-libgcc73-7.3.1-1.2.el7.centos.x86_64
scylla-libstdc++73-7.3.1-1.2.el7.centos.x86_64
scylla-manager-agent-2.2.2-0.20201223.32fed80f.x86_64
scylla-python3-3.8.5-0.20210104.24346215c2.x86_64
scylla-server-4.2.3-0.20210104.24346215c2.x86_64
scylla-tools-4.2.3-20210105.9d11db58267.noarch
scylla-tools-core-4.2.3-20210105.9d11db58267.noarch
```

Scylla Cluster - Backup Destination:
```
scylla-4.2.3-0.20210104.24346215c2.x86_64
scylla-conf-4.2.3-0.20210104.24346215c2.x86_64
scylla-debuginfo-4.2.3-0.20210104.24346215c2.x86_64
scylla-jmx-4.2.3-20210105.e351b009da.noarch
scylla-kernel-conf-4.2.3-0.20210104.24346215c2.x86_64
scylla-machine-image-4.2.3-20210105.ab5b881b69f.noarch
scylla-manager-agent-2.2.2-0.20201223.32fed80f.x86_64
scylla-python3-3.8.5-0.20210104.24346215c2.x86_64
scylla-server-4.2.3-0.20210104.24346215c2.x86_64
scylla-tools-4.2.3-20210105.9d11db58267.noarch
scylla-tools-core-4.2.3-20210105.9d11db58267.noarch
```

### sstableloader
**Command**: `./bin/scylla-sstableloader --verbose 4 --no-prepared -d '<list of ips>' /var/lib/scylla/data/temp/ks/cf/<location of sstables files>`

**Error:**
```
[...]

Issue #88, opened 2019-03-05
sstableloader null clustering column gives InvalidQueryException


elcallio:
It is (unfortunately) valid to have null/empty clustering values in a row, which manifests itself as empty value clustering prefixes in the cql row reader. 
However, translating this to value=null works poorly with bound/prepared statements, and there is no "clean" way to set an empty value. 

Options are to either exclude the value, which would be valid in the insert case, but not for an update. Or "fake" empty values by simply making the value the empty byte buffer value from sstable, and fake yet another codec into the driver that simply forwards the data raw.

elcallio:
See #57 for original problem, the solution to which uses null values. 
Those null values however do no longer work/do not work for some tables.

glommer:
we can't really skip it either in inserts because that row will have columns of its own, no ?

Can't we just not send a prepared statement in that case, if the problem is with bindings ?

elcallio:
I think it is binding in general that makes the null be really null. 
And no, inserts cannot really skip clustering either, though we could potentially have it unset. 
But I've sent a patch that simply changes the way these values are forwarded, so that we bypass the serialization stuff, and just send the raw data (i.e. "empty value") in the statement, which gives the desired results. 
Could potentially do this more to simplify UDT handling etc, but should really restructure the whole tool first before going there.

slivne:
patch "sstableloader: Use "empty" instead of null in bound values for cluster col"

Issue #57, opened 2018-01-25
sstableloader does not support udts


slivne:
scylla version 2.1.rc0-0.20180102.f371d1788
scylla-tools-java version 8efa88d932c70d37da63afc704560854b5d52ef9
cassandra version used 3.0.11

created the following schema

```
CREATE KEYSPACE keyspace1 WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;

CREATE TYPE keyspace1.typeone (
    value int,
    attributeone tinyint
);

CREATE TYPE keyspace1.typetwo (
    id text,
    attributetwo frozen<typeone>,
    attributethree frozen<typeone>
);

CREATE TABLE keyspace1.tableone (
    key int PRIMARY KEY,
    datetime timestamp,
    new_quotes map<text, frozen<list<frozen<typetwo>>>>,
    provider text
)

```

inserted the following data

```
cqlsh> insert INTO keyspace1.tableone (key,new_quotes,provider) values (1,{'a': [{id:'b', attributetwo: {value:1,attributeone:1}, attributethree:{value:1, attributeone:1}}]} ,'text');
cqlsh> select * from keyspace1.tableone ;

 key | datetime | new_quotes                                                                                                 | provider
-----+----------+------------------------------------------------------------------------------------------------------------+----------
   1 |     null | {'a': [{id: 'b', attributetwo: {value: 1, attributeone: 1}, attributethree: {value: 1, attributeone: 1}}]} |     text

```

tried to upload into scylla

```
sstableloader keyspace1/tableone-263562f001ed11e894dfa151b4bed413/mc-1-big-Data.db -v -x -d 127.0.0.1
CQL: 'INSERT INTO keyspace1.tableone (provider,key) values (?,?) USING TIMESTAMP ?;' [text, 1, 1516898386504595]
Preparing: INSERT INTO keyspace1.tableone (provider,key) values (?,?) USING TIMESTAMP ?;
CQL: 'UPDATE keyspace1.tableone USING TIMESTAMP ? SET new_quotes[?] = ? WHERE key = ?;' [1516898386504595, a, [java.nio.HeapByteBuffer[pos=8 lim=47 cap=47]], 1]
Preparing: UPDATE keyspace1.tableone USING TIMESTAMP ? SET new_quotes[?] = ? WHERE key = ?;
```

although there seem to be and insert for the provider,key and update for the new_quotes - nothing is found in scylla

```
[...]

PR #224, opened 2021-02-09
sstableloader: Use compound "where" restrictions for clustering

elcallio:
Refs #218

If we have multi-level clustering bounds (+1 column) with ranges,
we cannot use multiple where-clauses. Luckily, late insight (where
was this info 5 years ago?) tells us clustering restrictions are
legal to express as tuple expressions in CQL.

Note that this does not fully fix the above issue, as we still get
borked schema info from java driver due to case sensitive "desc"
comparison.

Note also that this is weakly tested since I cannot figure out
how to get dtest to accept and use sstableloader from a local
tools-java properly.

### DIFF
--- a/src/java/com/scylladb/tools/SSTableToCQL.java
+++ b/src/java/com/scylladb/tools/SSTableToCQL.java
@@ -23,6 +23,7 @@
 
 package com.scylladb.tools;
 
+import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
 import static java.util.concurrent.TimeUnit.MICROSECONDS;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
@@ -33,10 +34,12 @@ import static org.apache.cassandra.db.ClusteringBound.inclusiveStartOf;
 import java.nio.ByteBuffer;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
@@ -330,8 +333,9 @@ public class SSTableToCQL {
         long timestamp;
         int ttl;
         Multimap<ColumnDefinition, ColumnOp> values = MultimapBuilder.treeKeys().arrayListValues(1).build();
-        Multimap<ColumnDefinition, Pair<Comp, Object>> where = MultimapBuilder.treeKeys().arrayListValues(2).build();
-                
+
+        ClusteringWhere where;
+
         enum Comp {
             Equal("="),
             GreaterEqual(">="),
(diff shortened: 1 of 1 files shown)


elcallio:
nvm. managed to get dtest working and found bug.

elcallio:
updated.

src/java/com/scylladb/tools/SSTableToCQL.java
```diff
@@ -364,6 +368,66 @@ public Statistics getStatistics() {
...
+                this.lowBounds = new ArrayList<>(columnDefs.size());
+                this.highBounds = new ArrayList<>(columnDefs.size());
+
+                ClusteringPrefix spfx = start.clustering();
+                ClusteringPrefix epfx = end.clustering();
+
+                for (int i = 0; i < columnDefs.size(); i++) {
+                    ColumnDefinition column = columnDefs.get(i);
+                    if (i < spfx.size()) {
+                        lowBounds.add(column.cellValueType().compose(spfx.get(i)));
```
dekimir, line 388: I wonder if this will work properly with [reversed clustering order](http://docs.scylladb.com/getting-started/ddl/#reversing-the-clustering-order).  I expect that scrambles the relationship between `start` and `lowBounds`.

elcallio: No, this is handled in parsing/preprocessing earlier in the code (the latter for boundaries). In any case, the clustering object start/end are lo/hi according to clustering order. Which is why, earlier in the ref:ed issue, the problem of the java driver messing this clustering order up leads to other issues (the boundary logic getting messed up - we got sequential clustering bounds that became "lower", which is not supposed to happen).

dekimir: > No, this is handled in parsing/preprocessing earlier in the code (the latter for boundaries).

OK.  I hope we have tests covering that specifically.

> In any case, the clustering object start/end are lo/hi according to clustering order.

Yes, but not according to CQL `<` and `>`, which are on natural order.  Eg, an `int` clustering column doesn't change the meaning of `c<123` when its clustering order is reversed.  With natural order, `123` is the upper bound in storage, but with reversed order it's the lower bound, right?

elcallio: Ugh, that is a very good question. I sort of assumed the comparison followed sort order, esp. when using the tuple syntax....

dekimir: Not according to https://github.com/scylladb/scylla/issues/2050.

elcallio: Well, that is a bummer. That basically means we cannot use CQL syntax to express tombstones (or any other sstable CK restriction really) if sort order is reversed.

avikivity:
Let's put up some proposals for syntax.

1. magic reversed function

`(ck1, reversed(ck2), ck3) > ?`

Reverses the sense of ck2 when comparing. For it to work reversed has to be applied to DESC clustering columns.

2. clustering key compare function

`clustering_compare_gt((ck1, ck2, ck3), ?)`

Works only with the first input being a prefix of the clustering keys. Variants exist for all tuples that are prefixes of the clustering key, and with different comparisons (e.g. _le, _ge, _lt).

dekimir:
I dislike the magic reversed function because the explanation of its use is complicated.  Also, if tomorrow we add another ordering quirk, this function will no longer suffice.  I think it's better to have a comparator that sorts in clustering order, whatever it may be.

I'd avoid the "gt/lt" suffix because it conflicts with CQL `>` and `<`.  But I'm definitely in favour of something that spells out we're defining the CK range start and end (guaranteeing, of course, that start is before the end).

avikivity:
We can use cql DESC:

`(ck1, ck2 DESC, ck3) > (10, 3, 5)`

dekimir:
> We can use cql DESC:
> 
> `(ck1, ck2 DESC, ck3) > (10, 3, 5)`

That's basically the same as `reversed`.  In addition to having a complicated definition and solving too specific a problem, it has another disadvantage: you have to review the table schema to generate it correctly.  The other solution is schema-oblivious and thus more useful.

dekimir:
I propose `CK_STARTS_{WITH|AFTER}` and `CK_ENDS_{WITH|BEFORE}`, which take a tuple of values corresponding to a clustering prefix.  Eg:
```cql
SELECT * FROM t WHERE CK_STARTS_WITH(1,2,3) AND CK_ENDS_BEFORE(10,20)
SELECT * FROM t WHERE CK_STARTS_AFTER(1,2)
SELECT * FROM t WHERE CK_STARTS_WITH(1,2,3) AND CK_ENDS_WITH(10,20,30,40)
```

avikivity:
> I propose `CK_STARTS_{WITH|AFTER}` and `CK_ENDS_{WITH|BEFORE}`, which take a tuple of values corresponding to a clustering prefix. Eg:
> 
> ```sql
> SELECT * FROM t WHERE CK_STARTS_WITH(1,2,3) AND CK_ENDS_BEFORE(10,20)
> DELETE FROM t WHERE CK_STARTS_AFTER(1,2)
> DELETE FROM t WHERE CK_STARTS_WITH(1,2,3) AND CK_ENDS_WITH(10,20,30,40)
> ```

This maps directly to sstables, but is very different from SQL concepts which are oriented at expressing things in a mathematical way. 

Since it's intended to be used as a back door, not by users, that's acceptable. But use CLUSTERING_KEY_STARTS_*, not CK.

avikivity:
btw, we should minimize the number of query shapes we'll require. So maybe CLUSTERING_KEY_STARTS_WITH(?), where the input is a single tuple, with trailing NULLs for columns that are to be ignored. This limits us to 4 query shapes (open-ended range tombstones can be represented with a tuple of NULLs).

elcallio:
Why not just add a single keyword SCYLLA_USE_RAW_CLUSTERING (or similar) that just turns of the transformation of CQL-space restrictions into CK-space ones (since we are expressing CK already). Seems easier.

slivne:
@avikivity / @dekimir the change in scylla went in we can merge the sstableloader part

elcallio:
> 
> 
> Looks good, though it would be nice to know how this change affects #88 and #57. From the deleted code, I gather these issues required some special action that no longer seems to happen. Perhaps include an explanation in the commit message?

Both those are covered by this, since we are essentially able to simplify (and correctify) clustering expression. We're now just copying the clustering bounds from origin objects into the range expressions (except for insert/update, where we have singular expressions of equals).

src/java/com/scylladb/tools/SSTableToCQL.java
```diff
@@ -528,49 +549,100 @@ private void finish() {
...
+                        buf.append('(');
+
+                        i = 0;
+                        for (ColumnDefinition c : p.left.left) {
+                            if (i > 0) {
+                                buf.append(',');
+                            }
+                            String var = varName(c, instanceNumber);
+                            params.put(var, p.left.right.get(i));
+                            buf.append(':').append(var);
```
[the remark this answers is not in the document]
elcallio: ? This is neither the prepare, nor the use of it. 
Preparation and subsequent usage is done in the BulkLoader class, where we simply match the generated string against possibly previously prepared statements. We still have to compose the string for each cql row, but the server will skip parsing etc.

PR #229, opened 2021-02-26
cassandra-stress: fix failure due to the assert exception on disconnect when test is completed

dkropachev:
When test is completed c-s closes connection to the cluster.
due to the bug in driver it can fail there with assertException:

java.lang.AssertionError
        at com.datastax.driver.core.ConvictionPolicy$DefaultConvictionPolicy.signalConnectionClosed(ConvictionPolicy.java:90)
        at com.datastax.driver.core.Connection.closeAsync(Connection.java:953)
        at com.datastax.driver.core.HostConnectionPool.discardAvailableConnections(HostConnectionPool.java:882)
        at com.datastax.driver.core.HostConnectionPool.closeAsync(HostConnectionPool.java:843)

### DIFF
--- a/tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
+++ b/tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
@@ -256,6 +256,12 @@ public class JavaDriverClient
 
     public void disconnect()
     {
-        cluster.close();
+        try {
+            cluster.close();
+        } catch (Exception e) {
+            System.out.printf(
+                    "Failed to close connection due to the following error: %s",
+                    e.toString());
+        }
     }
 }


penberg:
@dkropachev The stack trace suggests this is a driver bug, no? Did upstream `cassandra-stress` have this issue as well? Let's /cc @elcallio too.

dkropachev:
> @dkropachev The stack trace suggests this is a driver bug, no? Did upstream `cassandra-stress` have this issue as well? Let's /cc @elcallio too.

@penberg , They have exactly the same code there, so potentially they can have the same issue, but since driver is different, it is possible that cassandra-driver does not pass exceptions through and does not trigger the failure.

dkropachev:
@penberg , can we have it merged ?

penberg:
@elcallio Can you please also review this?

dkropachev:
@penberg , @elcallio could you please review it and/or merge it ?

tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
```diff
@@ -256,6 +256,12 @@ public static ConsistencyLevel from(org.apache.cassandra.db.ConsistencyLevel cl)
...
     public void disconnect()
     {
-        cluster.close();
+        try {
+            cluster.close();
+        } catch (Exception e) {
+            System.out.printf(
+                    "Failed to close connection due to the following error: %s",
+                    e.toString());
+        }
```
elcallio, line 265: Is there no logger more appropriate for this?

elcallio (approved):
Catch + print stacktrace is the antothesis of java error handling, but sure, why not...

tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
penberg: @elcallio The reason I wanted you to review this is that it's not clear to me why `session.close()` even throws, and whether we want to propagate it from `disconnect` or not... A better logger would be nice.

elcallio: Given the stack trace above, and reading the code, I would say it is a bug somehow. If it is in atomicinteger or in the driver connection class I cannot say. Looking at the code it looks ok, it incements the connection count on open, and decrements it on close. Close checks that it is not called twice. Maybe if a connection is somehow created without the host counter being incremented? 
Is this running with our (sharded) driver?

dkropachev: Yes, it is running with sharded driver.
Let's address this issue in driver repo.
Meanwhile we can merge this PR since it is still valid, test should not fail just because it failed to close connection at the very end.

penberg: @elcallio There's no loggers in `cassandra-stress` AFAICT. @dkropachev let's switch to `System.err`, and I will merge this.

penberg:
@dkropachev I merged this as-is. Please check if we can switch to `System.err` and send an incremental fix if needed.

PR #232, opened 2021-03-09
cassandra.in.sh: Add path to rack/dc properties file to classpath

jul-stas:
Otherwise, when using `GossipingPropertyFileSnitch`, nodetool would not find it. On installed scylla this file is typically located in `$SCYLLA_CONF`, next to `scylla.yaml`.

Refs scylladb/scylla#7930

### DIFF
--- a/dist/common/cassandra.in.sh
+++ b/dist/common/cassandra.in.sh
@@ -27,7 +27,7 @@ for jar in /opt/scylladb/share/cassandra/*.jar; do
     CLASSPATH=$CLASSPATH:$jar
 done
 
-CLASSPATH="$CLASSPATH:$EXTRA_CLASSPATH"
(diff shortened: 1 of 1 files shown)


avikivity:
Cross-repository issues have different syntax: Ref scylladb/scylla#7930 (notice how your syntax isn't a hyperlink). "Ref" and not "Fixes" because it's the submodule update that actually fixes.

jul-stas:
> Cross-repository issues have different syntax: Ref [scylladb/scylla#7930](https://github.com/scylladb/scylla/issues/7930) (notice how your syntax isn't a hyperlink). "Ref" and not "Fixes" because it's the submodule update that actually fixes.

Fixed.
I knew abouth both... it was just a spark of retardation.

PR #237, opened 2021-03-30
fix: Add dummy implementation of getToppartitions

StarostaGit:
Normally we use the implementation found in the jmx repository.
This change adds a dummy implementation to the Cassandra implementation
file to fix build errors.


amnonh:
Not a bug but you can just return null, no one will ever going to call this method and if they do, getting a null back is fine

StarostaGit:
> Not a bug but you can just return null, no one will ever going to call this method and if they do, getting a null back is fine

Yes, that's way better
I forgot everything is nullable in Java

StarostaGit:
@amnonh done

amnonh:
LGTM

amnonh (approved):

PR #242, opened 2021-04-06
sstableloader: Handle non-prepared batches with ":" in identifier names

elcallio:
Because non-prepared batch statements cannot handle named variables,
we do variable replacement in this mode (terrible, yes). The regex
match does not deal with column/keyspace/table identifiers containing
":" (i.e. alternator).

Need to do boundary match (lookahead/lookbehind).

### DIFF
--- a/src/java/com/scylladb/tools/BulkLoader.java
+++ b/src/java/com/scylladb/tools/BulkLoader.java
@@ -798,7 +798,9 @@ public class BulkLoader {
             }
         }
 
-        private static final Pattern variable = Pattern.compile("\\:(\\w+)");
+        // #241 - cql table identifiers can have names with ":" in them.
+        // In that case, it should be escaped by now. Need to match against this.
+        private static final Pattern variable = Pattern.compile("(?<=[^\"]|^)\\:(\\w+)(?=[^\"]|$)");
 
         private void send(DecoratedKey key, long timestamp, String what, Map<String, Object> objects, boolean allowBatch) {
             SimpleStatement s;


nyh (commented):
I didn't understand the new regular expression, or what exactly it is supposed to do, so I don't know how to review this patch, unfortunately. So I care more about how you tested it:
How did you test that this change does what it needs to do - both that the classic cases (without a colon in column names) continues to work as it did before, and the unusual but legal case of a column name starting with a colon? Did you test it manually at least? 
Would be nice to have a unit test for sstableloader. But if we don't, at least the new dtests written by @Orenef11 should cover this bug - it should be trivial to him to add a third test case with "-nx". But before that, I was just curious how you @elcallio tested this patch.

elcallio:
By manually creating a table with a column ":i1" and reading an sstable from it. We don't have any unit testing in tools-java, så testing must be in dtests. I assumed the above failed tests could serve said purpose. 

The regex uses look-behind and look-ahead to ensure the presumed variable (:v1, :v2 etc) is not surrounded by `"` - i.e. either has a non-" char before/after or string start/end.

nyh:
> By manually creating a table with a column ":i1" and reading an sstable from it. We don't have any unit testing in tools-java, så testing must be in dtests. I assumed the above failed tests could serve said purpose.
> 
> The regex uses look-behind and look-ahead to ensure the presumed variable (:v1, :v2 etc) is not surrounded by `"` - i.e. either has a non-" char before/after or string start/end.

What if the column name includes a (quoted) quote character? :-)
But maybe I'm asking for too much. If you think this is good enough like this, I can commit it.

elcallio:
> 
> 
> > By manually creating a table with a column ":i1" and reading an sstable from it. We don't have any unit testing in tools-java, så testing must be in dtests. I assumed the above failed tests could serve said purpose.
> > The regex uses look-behind and look-ahead to ensure the presumed variable (:v1, :v2 etc) is not surrounded by `"` - i.e. either has a non-" char before/after or string start/end.
> 
> What if the column name includes a (quoted) quote character? :-)
> But maybe I'm asking for too much. If you think this is good enough like this, I can commit it.

We are not looking for column names, in fact we are trying to bypass them and only find the variables. Since `:v1`, `:v2` etc  are not valid unquotes column/table/keyspace names, there must always be quotes in the CQL string iff a column has an overlapping name. Thus we can match them out by checking we have a "raw" `:<something>`.

PR #250, opened 2021-05-10
toppartitions: Fix toppartitions to only jmx once

StarostaGit:
In line with previous API, nodetool called jmx's toppartitions
for each sampler separately. A better way to do this is to call
it once and then pick results for the samplers we need.


src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -673,7 +673,7 @@
     /** Sets the hinted handoff throttle in kb per second, per delivery thread. */
     public void setHintedHandoffThrottleInKB(int throttleInKB);
 
-    public CompositeData getToppartitions(String sampler, List<String> keyspaceFilters, List<String> tableFilters, int duration, int capacity, int count) throws OpenDataException;
+    public Map<String, CompositeData> getToppartitions(List<String> keyspaceFilters, List<String> tableFilters, int duration, int capacity, int count) throws OpenDataException;
```
avikivity, line 676: How does this work? You changed the interface, but not the implementation?

Maybe the implementation is autogenerated, but it has to be generated from some IDL?

StarostaGit: Implementation is in the jmx PR
meant to mark it as draft, sorry
the whole fix is composed of three PRs, we'll link them here once I describe the commits in the third one

StarostaGit: But to answer the question more thoroughly, nodetool uses this interface which searches for the implementation of it in the jmx package

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -482,11 +482,12 @@ public void repairAsync(final PrintStream out, final String keyspace, Map<String
 
     public Map<Sampler, CompositeData> getToppartitions(List<String> keyspaceFilters, List<String> tableFilters, int capacity, int duration, int count, List<Sampler> samplers) throws OpenDataException
     {
-        Map<Sampler, CompositeData> result = Maps.newHashMap();
+        Map<String, CompositeData> toppartitions = ssProxy.getToppartitions(keyspaceFilters, tableFilters, duration, capacity, count);
```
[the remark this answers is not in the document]
StarostaGit: The samplers are passed into this method, but not propagated further, contrary to how it worked before. The reason is we want to only call the jmx and therefore the scylla backend once (since it always gives us info for all samplers anyway) and only here, in the NodeProbe code, extract counters for the samplers the caller requested

src/java/org/apache/cassandra/service/StorageServiceMBean.java
avikivity: Then how do upgrades work? We'll need to make the old/new versions interoperate.

StarostaGit: I've updated the PR desc
scylladb/scylla-jmx#165 has to be merged first into scylla-jmx, only then this can be merged as well

StarostaGit:
@avikivity You wanted me to make this work with older version of the JMX call. Asked how old under the previous PR, but maybe it went unnoticed so reposting here

avikivity:
> @avikivity You wanted me to make this work with older version of the JMX call. Asked how old under the previous PR, but maybe it went unnoticed so reposting here

I am quite lossy, ping when needed. It's annoying and I apologize.

by "older" I meant the previous version before your change.

> 
> I already made it detect the absence of the new method, just have to know which method to redirect to - before this set of PRs or before the generic toppartitions API

Isn't that sufficient? It will use the new API is present and fall back to the old one if not.

StarostaGit:
no worries

> > @avikivity
> by "older" I meant the previous version before your change.

That's what I wanted to know, great, will post the change soon

StarostaGit:
@avikivity done, it falls back successfully when the new method is not found

bhalevy:
@StarostaGit I still see 2 calls in the scylla log, e.g.:
```
INFO  2021-06-16 10:25:07,157 [shard 0] api - toppartitions query: #table_filters=all #keyspace_filters=all duration=10000 list_size= capacity=256
INFO  2021-06-16 10:25:17,174 [shard 0] api - toppartitions query: #table_filters=all #keyspace_filters=all duration=10000 list_size= capacity=256
```

I'm not exactly sure what this PR was trying to achieve but the second calls overrides the first one, i.e. nodetool toppartitions returns the stats only from the 2nd period as far as I can tell from the dtest.

bhalevy:
Cc @haaawk let's get the toppartitions dtests pass.
They are failing since https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/781/testReport/toppartitions_test/TestTopPartitions/test_any_of_required_parameters_is_missing/

StarostaGit:
> @StarostaGit I still see 2 calls in the scylla log, e.g.:
> 
> ```
> INFO  2021-06-16 10:25:07,157 [shard 0] api - toppartitions query: #table_filters=all #keyspace_filters=all duration=10000 list_size= capacity=256
> INFO  2021-06-16 10:25:17,174 [shard 0] api - toppartitions query: #table_filters=all #keyspace_filters=all duration=10000 list_size= capacity=256
> ```
> 
> I'm not exactly sure what this PR was trying to achieve but the second calls overrides the first one, i.e. nodetool toppartitions returns the stats only from the 2nd period as far as I can tell from the dtest.

It had a wrong Class parameter passed for lists in the older/newer method call dispatch, hence for new ones it was still 2 calls to the method. There was a followup for that in #256 and it should be ok once it went in
Same for the dtests, let me call the BYO again

bhalevy:
> It had a wrong Class parameter passed for lists in the older/newer method call dispatch, hence for new ones it was still 2 calls to the method. There was a followup for that in #256 and it should be ok once it went in
> Same for the dtests, let me call the BYO again

fwiw, they pass for me locally with #256

bhalevy:
thanks

StarostaGit:
it works locally for me as well, the PR with updated dtests is also just waiting for #256 so that the BYO job passes

bhalevy:
what PR updating dtests?

StarostaGit:
> what PR updating dtests? [scylladb/scylla-dtest#2147](https://github.com/scylladb/scylla-dtest/pull/2147)?

Yes, this one

PR #252, opened 2021-05-14
cassandra-stress: Remove maxPendingPerConnection default

haaawk:
cassandra-stress uses 128 as a default for maxPendingPerConnection.
This is very wrong. 128 used to be the default value in the driver
up to the version 3 of the protocol. Since V3 the default is 1024.

The behaviour of cassandra-stress has caused driver to throw
BusyPoolException because each connection was limited to only 128
inflight requests. For more info see:
https://github.com/scylladb/scylla/issues/7748

This patch removes the default maxPendingPerConnection value from
cassandra-stress. This way driver will decide on its own what's the
appropriate default depending on the version of the CQL protocol used.

Fixes https://github.com/scylladb/scylla/issues/7748

### DIFF
--- a/tools/stress/src/org/apache/cassandra/stress/settings/SettingsMode.java
+++ b/tools/stress/src/org/apache/cassandra/stress/settings/SettingsMode.java
@@ -164,7 +164,7 @@ public class SettingsMode implements Serializable
         final OptionSimple user = new OptionSimple("user=", ".+", null, "username", false);
         final OptionSimple password = new OptionSimple("password=", ".+", null, "password", false);
         final OptionSimple authProvider = new OptionSimple("auth-provider=", ".*", null, "Fully qualified implementation of com.datastax.driver.core.AuthProvider", false);
-        final OptionSimple maxPendingPerConnection = new OptionSimple("maxPending=", "[0-9]+", "128", "Maximum pending requests per connection", false);
+        final OptionSimple maxPendingPerConnection = new OptionSimple("maxPending=", "[0-9]+", null, "Maximum pending requests per connection", false);
         final OptionSimple connectionsPerHost = new OptionSimple("connectionsPerHost=", "[0-9]+", "8", "Number of connections per host", false);
 
         abstract OptionSimple mode();
--- a/tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
+++ b/tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
@@ -46,7 +46,7 @@ public class JavaDriverClient
     public final String username;
     public final String password;
     public final AuthProvider authProvider;
-    public final int maxPendingPerConnection;
+    public final Integer maxPendingPerConnection;
     public final int connectionsPerHost;
 
     private final ProtocolVersion protocolVersion;
@@ -84,7 +84,7 @@ public class JavaDriverClient
         //See https://issues.apache.org/jira/browse/CASSANDRA-7217
         int requestsPerConnection = (maxThreadCount / connectionsPerHost) + connectionsPerHost;
 
-        maxPendingPerConnection = settings.mode.maxPendingPerConnection == null ? Math.max(128, requestsPerConnection ) : settings.mode.maxPendingPerConnection;
+        maxPendingPerConnection = settings.mode.maxPendingPerConnection;
     }
 
     private LoadBalancingPolicy loadBalancingPolicy(StressSettings settings)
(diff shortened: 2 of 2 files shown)


haaawk:
@scylladb/scylla-tools-maint please have a look

avelanarius:
It would be great if we confirmed that it solves https://github.com/scylladb/scylla/issues/7748 in the same longevity test as originally reported (cc: @ShlomiBalalis - just adding `maxPending=1024` to `cassandra-stress` achieves the same as this fix) - I have tested this PR locally and it fixed the problem, but it would be nice to confirm on the actual test.

haaawk:
@ShlomiBalalis ping any luck with re-running the test?

penberg:
@haaawk What does upstream `cassandra-stress` do? I am always uneasy about changing behaviour because you get different results depending on which version of the tools you used. Test code could pass the larger value, no?

haaawk:
It seems that they have the same change in upstream @penberg -> https://github.com/apache/cassandra/blob/trunk/tools/stress/src/org/apache/cassandra/stress/settings/SettingsMode.java#L68

haaawk:
> It seems that they have the same change in upstream @penberg -> https://github.com/apache/cassandra/blob/trunk/tools/stress/src/org/apache/cassandra/stress/settings/SettingsMode.java#L68

Actually no -> https://github.com/apache/cassandra/blob/trunk/tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java#L93

They still limit do 128 by default. We could try to upstream the change as it makes sense considering defaults in modern drivers.

haaawk:
> > It seems that they have the same change in upstream @penberg -> https://github.com/apache/cassandra/blob/trunk/tools/stress/src/org/apache/cassandra/stress/settings/SettingsMode.java#L68
> 
> Actually no -> https://github.com/apache/cassandra/blob/trunk/tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java#L93
> 
> They still limit do 128 by default. We could try to upstream the change as it makes sense considering defaults in modern drivers.

It seems that the limit is there since 2015 and since then was never updated -> https://github.com/apache/cassandra/commit/a0076e70eed5501ac9d8c3ff41ce8018710a1585#diff-0889dbee1bb3ebfbb9cd0d40c96c538f3b50887951728bf59ae7e58f6b6d6c70

This is what happens when you copy default values from a library and then assume it will never change :/

haaawk:
> @haaawk What does upstream `cassandra-stress` do? I am always uneasy about changing behaviour because you get different results depending on which version of the tools you used. Test code could pass the larger value, no?

I understand your concern and it is true that this can be configured with command line parameter. The question is -> Are you willing to educate all users and scylla developers/QA and handle every case when someone forgets to set the param and creates an issue? In my opinion the only way to handle this in a scalable way is to fix the default. Otherwise we will be spending time on this over and over again.

haaawk:
> @haaawk What does upstream `cassandra-stress` do? I am always uneasy about changing behaviour because you get different results depending on which version of the tools you used. Test code could pass the larger value, no?

I guess my final argument would be that you will already get much different results depending on which version of cassandra-stress you use because ours uses shard-aware driver ;)

haaawk:
ping

haaawk:
@avikivity @penberg @scylladb/scylla-tools-maint pingus maximus

amoskong:
@haaawk @slivne  do we want to backport this fix to 4.5?

/Cc @roydahan

haaawk:
@amoskong please keep in mind there's a follow-up that fixes a bug this PR has: https://github.com/scylladb/scylla-tools-java/pull/258

avikivity:
@haaawk do we want to backport it? How likely is a user to experience it?

haaawk:
@avikivity everyone who does not set the parameter will be subject to a wrong default.

avikivity:
What's the impact of the wrong default?

haaawk:
Number of concurrent requests per connection is limited to 128. For some context, old driver had 128 as a default and that was copyied to c-s but later newer drivers raised the limit to 1024 and c-s was never updated keeping way to low limit. This is especially a problem with shard awarness because previously each reaquest could go through any connection and with shard awareness it goes through a selected connection so it's much easier to saturate the limit.

amoskong:
Hi @avikivity so we won't backport the fix to 4.5?  If so we will set a `maxPending=1024` for cs workloads in SCT.

bhalevy:
I think it's a good idea to backport to 4.5 to simplify SCT.

PR #255, opened 2021-05-26
reloc: add arch to relocatable package filename

syuu1228:
Add architecture name for relocatable packages, to support distributing
both x86_64 version and aarch64 version.

See scylladb/scylla#8675


syuu1228:
No need to rename noarch packages, so closing this https://github.com/scylladb/scylla/pull/8709#issuecomment-863520892

avikivity:
Not sure if the embedded python driver for cqlsh is really noarch.

PR #256, opened 2021-06-03
node_probe: toppartitions: Fix wrong class in getMethod

StarostaGit:
We check for the existence of the new toppartitions JMX API
using Class.getMethod, but use wrong class for List<String>
in parameter list. This change fixes that.

### DIFF
--- a/src/java/org/apache/cassandra/tools/NodeProbe.java
+++ b/src/java/org/apache/cassandra/tools/NodeProbe.java
@@ -488,7 +488,7 @@ public class NodeProbe implements AutoCloseable
         try
         {
             Class[] cArg = new Class[6];
-            cArg[0] = cArg[1] = cArg[2] = (Class<List<String>>) Collections.<String>emptyList().getClass();
+            cArg[0] = cArg[1] = cArg[2] = List.class;
             cArg[3] = cArg[4] = cArg[5] = int.class;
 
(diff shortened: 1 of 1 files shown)


amnonh:
typically, nodeprobe is taken from Cassandra as-is, why is it different in this case?

StarostaGit:
As is? Isn't `NodeProbe.java`our file? I changed this code throughout multiple PRs. And this is an ammendment to one of those

amnonh:
for a very long time, nodeprob and nodetool in general were Cassandra compatible and while this rule no longer holds, there should be a reason why it's not.

StarostaGit:
> No, nodeprobe connect to scylla-jmx.
> for a very long time, nodeprob and nodetool in general were Cassandra compatible and while this rule no longer holds, there should be a reason why it's not.
> Is this something that got broken because of other change?

Yes, it's the method introduced to handle the generic toppartitions calls, #250 introduced the `try...catch` for legacy calls and I forgot to check that when both are present the newer one gets called. And the bug was because of the wrong Class object passed to `getMethod`

amnonh:
LGTM

amnonh (approved):

bhalevy:
@avikivity can you please update the tools/java submodule in scylla to include this fix?

bhalevy:
Also @scylladb/scylla-maint (doesn't have to be Avi.... :))

PR #261, opened 2021-06-16
reloc: increment relocatable package version

syuu1228:
This is needed for moving unified package contents to sub-directory.
This package itself has no change.


fruch:
@syuu1228 is this still needed ?

PR #262, opened 2021-06-25
install.sh: add supervisor support

syuu1228:
Bring supervisor support from dist/docker to install.sh, make it
installable from relocatable package.
This enables to use supervisor with nonroot / offline environment,
and also make relocatable package able to run in Docker environment.

See scylladb/scylla#8849


syuu1228:
Closing, we decided not changing this module. See: scylladb/scylla#9142

PR #267, opened 2021-07-18
dist/redhat: change PyYAML filepath to allow installing on SLES15

syuu1228:
It seems like PyYAML package on SLES15 does not have _yaml.so,
we need to use other file to specify the package.

Fixes scylladb/scylla#9045


amoskong (approved):
Looks good.

PR #269, opened 2021-07-26
nodetool: add triggeroffstrategycompaction

bhalevy:
Invoke new api to trigger offstrategy compaction
on keyspaces/tables.

Depends on https://github.com/scylladb/scylla-jmx/pull/172 and (indirectly) on scylladb/scylla#9095


penberg:
@bhalevy @tzach Would not adding a flag to `nodetool compact` make more sense for discoverability?

asias:
> @bhalevy @tzach Would not adding a flag to `nodetool compact` make more sense for discoverability?

I think so. In most of the cases, user will not call nodetool triggeroffstrategycompaction at all. The cluster will trigger it automatically.  In practice, the 5 minutes automatic timeout is supposed to be much less than the repair time.

raphaelsc:
On Fri, Jul 30, 2021 at 8:36 AM Asias He ***@***.***> wrote:

> @bhalevy <https://github.com/bhalevy> @tzach <https://github.com/tzach>
> Would not adding a flag to nodetool compact make more sense for
> discoverability?
>
> I think so. In most of the cases, user will not call nodetool
> triggeroffstrategycompaction at all. The cluster will trigger it
> automatically. In practice, the 5 minutes automatic timeout is supposed to
> be much less than the repair time.
>
agreed. better to add an --off-strategy flag instead.

Issue #268, opened 2021-07-21
nodetool scrub: add support for segregate mode


denesb:
Added by https://github.com/scylladb/scylla/commit/34643ac99 and https://github.com/scylladb/scylla/commit/550a1cd03.

bhalevy:
Fixed by #272

PR #270, opened 2021-07-27
nodetool: support validate

### LINKED ISSUE #263: nodetool: add method to invoke validation compaction
Scylla issue for scylla implementation: https://github.com/scylladb/scylla/issues/7736

bhalevy:
Add support for `nodetool validate [ks [cf...]]`
Based on the api added in scylladb/scylla@b0ef57c83390ada3152568f06769ad76c596cb5e

Depends on https://github.com/scylladb/scylla-jmx/pull/173

Fixes #263

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -2983,6 +2983,16 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
         return status.statusCode;
     }
 
+    public int validate(String keyspaceName, String... tables) throws IOException, ExecutionException, InterruptedException
+    {
+        return validate(0, keyspaceName, tables);
+    }
+
+    public int validate(int jobs, String keyspaceName, String... tables) throws IOException, ExecutionException, InterruptedException
+    {
+        throw new RuntimeException("Validate is not implemented");
+    }
+
     public int scrub(boolean disableSnapshot, boolean skipCorrupted, String keyspaceName, String... tables) throws IOException, ExecutionException, InterruptedException
     {
         return scrub(disableSnapshot, skipCorrupted, true, 0, keyspaceName, tables);
--- a/src/java/org/apache/cassandra/service/StorageServiceMBean.java
+++ b/src/java/org/apache/cassandra/service/StorageServiceMBean.java
@@ -270,6 +270,19 @@ public interface StorageServiceMBean extends NotificationEmitter
     public int forceKeyspaceCleanup(int jobs, String keyspaceName, String... tables) throws IOException, ExecutionException, InterruptedException;
 
     /**
+     * Run validation compaction of tables in a single keyspace.
+     * If tables array is empty, all tables are validated.
+     *
+     * This is essentially a read-only version of scrub.
+     */
+    @Deprecated
+    public int validate(String keyspaceName, String... tables)
+            throws IOException, ExecutionException, InterruptedException;
+
+    public int validate(int jobs, String keyspaceName, String... tables)
+            throws IOException, ExecutionException, InterruptedException;
+
+    /**
      * Scrub (deserialize + reserialize at the latest version, skipping bad rows if any) the given keyspace.
      * If tableNames array is empty, all CFs are scrubbed.
      *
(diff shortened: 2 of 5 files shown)


bhalevy:
Cc @raphaelsc

raphaelsc:
@bhalevy isn't it better to instead extend nodetool scrub with a --dry-run option? so we won't have to come up with yet another tool.

bhalevy:
> @bhalevy isn't it better to instead extend nodetool scrub with a --dry-run option? so we won't have to come up with yet another tool.

It's an interesting approach.

@denesb already opened #268 to support a new, `segregate` scrub mode.
I think it's going to be easier to represent `validate` as an additional scrub mode, rather than adding a `--dry-run` option
See scylladb/scylla@34643ac997.

This will require changing the api that was added in scylladb/scylla@b0ef57c833

denesb:
> @bhalevy isn't it better to instead extend nodetool scrub with a --dry-run option? so we won't have to come up with yet another tool.

I think this would open up a lot of confusion about what `--dry-run` means when any of the other flags are set to any value. IMO "validate" is just a more straightforward name and it also better describes what this really is.

raphaelsc:
On Thu, Jul 29, 2021 at 3:43 AM Botond Dénes ***@***.***>
wrote:

> @bhalevy <https://github.com/bhalevy> isn't it better to instead extend
> nodetool scrub with a --dry-run option? so we won't have to come up with
> yet another tool.
>
> I think this would open up a lot of confusion about what --dry-run means
> when any of the other flags are set to any value. IMO "validate" is just a
> more straightforward name and it also better describes what this really is.
>
Just realized that validation compaction in C* has a different meaning
"During a repair, each replica will build a merkle tree, using what is
called a “validation compaction”. It is basically a compaction without the
write phase, the output being a tree of hashes.". So I think we better
change the name to avoid confusion. What an ambiguous name they come up
with. Not the first though :-) See cleanup command, for example, which is
only intended to remove non-owned data, and received a generic name
misleading users into thinking that the command clean up other things as
well.

With --dry-run, the only difference is that scrub becomes read-only. If one
needs only a binary response, nodetool scrub --dry-run will work. If you
one needs full scope on corruption, then nodetool scrub
--dry-run --skip-corrupted will do it. Default behavior is to validate all
columns.

denesb:
>> On Thu, Jul 29, 2021 at 3:43 AM Botond Dénes ***@***.***> wrote: @bhalevy <https://github.com/bhalevy> isn't it better to instead extend nodetool scrub with a --dry-run option? so we won't have to come up with yet another tool. I think this would open up a lot of confusion about what --dry-run means when any of the other flags are set to any value. IMO "validate" is just a more straightforward name and it also better describes what this really is.
>
> Just realized that validation compaction in C* has a different meaning "During a repair, each replica will build a merkle tree, using what is called a “validation compaction”. It is basically a compaction without the write phase, the output being a tree of hashes.". So I think we better change the name to avoid confusion.

So this was an internal thing, not something user facing, right? If so users are not likely to even know about it and thus be confused.  We don't need to be unambiguous to C* developers :).

raphaelsc:
On Fri, Jul 30, 2021 at 2:45 AM Botond Dénes ***@***.***>
wrote:

> On Thu, Jul 29, 2021 at 3:43 AM Botond Dénes *@*.***> wrote: @bhalevy
> <https://github.com/bhalevy> https://github.com/bhalevy isn't it better
> to instead extend nodetool scrub with a --dry-run option? so we won't have
> to come up with yet another tool. I think this would open up a lot of
> confusion about what --dry-run means when any of the other flags are set to
> any value. IMO "validate" is just a more straightforward name and it also
> better describes what this really is.
>
> Just realized that validation compaction in C* has a different meaning
> "During a repair, each replica will build a merkle tree, using what is
> called a “validation compaction”. It is basically a compaction without the
> write phase, the output being a tree of hashes.". So I think we better
> change the name to avoid confusion.
>
> So this was an internal thing, not something user facing, right? If so
> users are not likely to even know about it and thus be confused. We don't
> need to be unambiguous to C* developers :).
>
VALIDATION will show up in nodetool compactionstats. So there will be this
difference which users will face. C* users will be confused because they
use compactionstats to track validation compactions as defined above.
Perhaps rename the validation string to something more specific to avoid
any sort of confusion?

denesb:
> VALIDATION will show up in nodetool compactionstats. So there will be this difference which users will face. C* users will be confused because they use compactionstats to track validation compactions as defined above. Perhaps rename the validation string to something more specific to avoid any sort of confusion?

I think at that point it is just better to go with `scrub --dry-mode`. @penberg also noted that `nodetool validate` is not present in C*'s nodetool.

bhalevy:
Functionality achieved by https://github.com/scylladb/scylla-tools-java/pull/272 as `nodetool scrub --mode VALIDATE`

Issue #251, opened 2021-05-11
Add '--ignore-dead-nodes' option support to nodetool removenode


asias:
Since commit

```
commit 829b4c14380020fa4058335c4c43cefec135b3de
Author: Asias He <asias@scylladb.com>
Date:   Mon Nov 16 16:13:44 2020 +0800

    repair: Make removenode safe by default
```

Scylla added the ability to exclude dead nodes during removenode operation.

If the user want the removenode operation to succeed even if some of the nodes
are not available, the user has to explicitly pass a list of nodes that can be
skipped for the operation.

The nodetool option can be:
```
$ nodetool removenode --ignore-dead-nodes <list_of_dead_nodes_to_ignore> <host_id>
````

Example restful api:
    
```
$ curl -X POST "http://127.0.0.1:10000/storage_service/remove_node/?host_id=7bd303e9-4c7b-4915-84f6-343d0dbd9a49&ignore_nodes=127.0.0.3,127.0.0.5"
    
```

**Comment by tarzanek** [comment]
it seems https://github.com/scylladb/scylla-jmx/pull/178/files adds support for option
but it doesn't really pass on the option to the REST API

```
[root@7cde2899a85b /]# nodetool removenode --ignore-dead-nodes 172.17.0.3,172.17.0.4 f9a3b363-f2dd-4ff0-a86b-27d4294e2983
nodetool: Scylla API server HTTP POST to URL '/storage_service/remove_node' failed: std::runtime_error (removenode[b31b7293-d7d9-4b41-b4cb-57c3d91fa72b]: Nodes={172.17.0.4} needed for removenode operation are down. It is highly recommended to fix the down nodes and try again. To proceed with best-effort mode which might cause data inconsistency, run nodetool removenode --ignore-dead-nodes <list_of_dead_nodes> <host_id>. E.g., nodetool removenode --ignore-dead-nodes 127.0.0.1,127.0.0.2 817e9515-316f-4fe3-aaab-b00d6f12dddd)
```

**Comment by tarzanek** [comment]
while direct
```
curl -X POST "http://127.0.0.1:10000/storage_service/remove_node/?host_id=f9a3b363-f2dd-4ff0-a86b-27d4294e2983&ignore_nodes=172.17.0.4,172.17.0.3" 
```
worked 
(found by @fee-mendes  )

bhalevy:
> was that option tested with multiple ignore nodes IPs ? @bhalevy ?

Yes. It's being tested regularly.
[...]

PR #278, opened 2021-10-01
removeNode: support ignoreNodes options

### LINKED ISSUE #225: nodetool removenode doesn't support "--ignore-dead-nodes" option
`--ignore-dead-nodes` option was added to 'removenode` in https://github.com/scylladb/scylla/pull/7626 and works via API, but it wasn't enabled in nodetool

bhalevy:
Fixes #225

DTest:
```
DTEST_REQUIRE=disabled CASSANDRA_DIR=../scylla/build/dev NOSE_PROCESSES=2 ./scripts/run_test.sh update_cluster_layout_tests.py:TestUpdateClusterLayout.{simple_removenode_1_test,simple_removenode_2_test,simple_removenode_3_test,simple_remove_2_nodes_3_test,simple_removenode_3_nodetool_test,simple_remove_2_nodes_3_nodetool_test,simple_removenode_4_test,simple_removenode_5_test}
```

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -4395,6 +4395,14 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
         removingNode = null;
     }
 
+    public void removeNode(String hostIdString, String ignoreNodes)
+    {
+        if (ignoreNodes != null)
+            throw new UnsupportedOperationException("ignoreNodes is not supported.");
+
+        removeNode(hostIdString);
+    }
+
     public void confirmReplication(InetAddress node)
     {
         // replicatingNodes can be empty in the case where this node used to be a removal coordinator,
--- a/src/java/org/apache/cassandra/tools/nodetool/RemoveNode.java
+++ b/src/java/org/apache/cassandra/tools/nodetool/RemoveNode.java
@@ -20,6 +20,7 @@ package org.apache.cassandra.tools.nodetool;
 import static org.apache.commons.lang3.StringUtils.EMPTY;
 import io.airlift.command.Arguments;
 import io.airlift.command.Command;
+import io.airlift.command.Option;
 
 import org.apache.cassandra.tools.NodeProbe;
 import org.apache.cassandra.tools.NodeTool.NodeToolCmd;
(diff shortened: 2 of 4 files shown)


bhalevy:
Depends on https://github.com/scylladb/scylla-jmx/pull/178

bhalevy:
@asias ping review

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4395,6 +4395,14 @@ public void removeNode(String hostIdString)
         removingNode = null;
     }
 
+    public void removeNode(String hostIdString, String ignoreNodes)
+    {
+        if (ignoreNodes != null)
+            throw new UnsupportedOperationException("ignoreNodes is not supported.");
```
haaawk, line 4401: Nit: The convention in Java is to always add {} block after `if`

haaawk: I see that this codebase does not exactly follow the convention so it might be ok to skip {} block here.

haaawk (approved):

src/java/org/apache/cassandra/service/StorageService.java
bhalevy: > I see that this codebase does not exactly follow the convention so it might be ok to skip {} block here.

Right.  I just stuck to the existing convention in the file.

bhalevy:
@scylladb/scylla-maint please merge after https://github.com/scylladb/scylla-jmx/pull/178 and scylla submodule update

src/java/org/apache/cassandra/tools/nodetool/RemoveNode.java
```diff
@@ -30,6 +31,9 @@
     @Arguments(title = "remove_operation", usage = "<status>|<force>|<ID>", description = "Show status of current node removal, force completion of pending removal, or remove provided ID", required = true)
     private String removeOperation = EMPTY;
 
+    @Option(title = "ignore_dead_nodes", name = {"-ignore", "--ignore-nodes", "--ignore-dead-nodes"}, description = "Use -ignore to specify a comma-separated list of dead nodes to ignore during removenode")
```
asias, line 34: @bhalevy in your implementation, do we support 3 options to specify the same?

I wish we only had one 

nodetool removenode --ignore-dead-nodes <list_of_dead_nodes_to_ignore> <host_id>

like I mentioned in https://github.com/scylladb/scylla-tools-java/issues/251

bhalevy: @asias the multiple options are only at the java level, the api still supports only one

asias: Yes, I know you did not change the restful API. I hope we only have a single option --ignore-dead-nodes for this, not -ignore and --ignore-nodes. We want to stress that we are ignoring dead node, not any nodes in the cluster. We need to use this option very careful, so short alias does not make much sense. The more options, the more to  document  and more possible confusion.

bhalevy: Feel free to submit a fix

asias: Cool. I send a PR here: https://github.com/scylladb/scylla-tools-java/pull/286

PR #285, opened 2022-01-08
dist/debian:set either python (>=2.7) or python2

yaronkaikov:
The changes we did in https://github.com/scylladb/scylla-tools-java/commit/6249bfbe2ff9f4be5ddc6cd3af6f191a636784eb breaks scylla installation in debian9, ubuntu16.04 and ubuntu18.04

Fixes: scylladb/scylla#9886

### DIFF
--- a/dist/debian/control.template
+++ b/dist/debian/control.template
@@ -17,7 +17,7 @@ Description: Scylla database tools
 
 Package: %{product}-tools-core
 Architecture: all
-Depends: openjdk-8-jre-headless | openjdk-8-jre | oracle-java8-set-default | adoptopenjdk-8-hotspot-jre | openjdk-11-jre-headless | openjdk-11-jre | oracle-java11-set-default, python2, procps
+Depends: openjdk-8-jre-headless | openjdk-8-jre | oracle-java8-set-default | adoptopenjdk-8-hotspot-jre | openjdk-11-jre-headless | openjdk-11-jre | oracle-java11-set-default, python (>=2.7) | python2, procps
 Description: Scylla database tools core files
  Core files for scylla database tools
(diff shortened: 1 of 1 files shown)


dist/debian/control.template
```diff
@@ -17,7 +17,7 @@ Description: Scylla database tools
 
 Package: %{product}-tools-core
 Architecture: all
-Depends: openjdk-8-jre-headless | openjdk-8-jre | oracle-java8-set-default | adoptopenjdk-8-hotspot-jre | openjdk-11-jre-headless | openjdk-11-jre | oracle-java11-set-default, python2, procps
+Depends: openjdk-8-jre-headless | openjdk-8-jre | oracle-java8-set-default | adoptopenjdk-8-hotspot-jre | openjdk-11-jre-headless | openjdk-11-jre | oracle-java11-set-default, python (>=2.7) | python2, procps
```
nyh, line 20: Did you test this also in the Ubuntu versions that commit 6249bfb was written for?

That commit suggested that having `python (>=2.7)` wasn't enough, because that could get you python 3 while we need python2. So - won't we have the same problem again after this patch? Or maybe I'm misunderstanding how the  `|` works?

yaronkaikov: Yes, i created the packages locally . then installed the packages manually.

So the issue is, for debian9 for example. once you install `python2.7` (which i guess it required by tools-java). and type the command `whereis python` you will get: `python: /usr/bin/python2.7 /usr/lib/python2.7 /etc/python2.7 /usr/local/lib/python2.7`

While for ubuntu:22.04 after we install `python2.7` package `whereis python` will give us `python:`

PR #297, opened 2022-05-24
Sync with Cassandra 3.11.13 and update a few dependencies

avelanarius:
This PR consists of my 4 commits (GitHub shows all commits from merging upstream branch...):

The first commit `Sync with Cassandra 3.11.13` merges the cassandra-3.11 branch 
from upstream Cassandra repository at the last commit of Cassandra 3.11.13. It also deals 
with the only "real conflict" due to the merge: new method `maybeWarnOnOptions()` had
to be implemented in one of our own classes (`EverywhereStrategy`).

The second commit: `Remove lib/ directory and its JAR files` deals with the fact that
after 2fdf1d27620fe2bce2a3c5989beab40b05fda1e6 appeared in our repository,
dependency management is no longer handled by putting JAR blobs to `lib/`. Instead,
it's fetched dynamically from Maven and put in that folder at the build time. The commit
from upstream repository did remove this directory, but this commit was only partially merged
before to our fork.

Having this folder present in the repo means that some older 
dependencies will linger around - new dependency handling will just
add new JAR files (newer version of dependenecy) and leave the old one
unchanged. This could cause warnings of security scanners.

The third commit `Update version of Netty dependency`. This commit was only present
on Cassandra 4.x branch (not 3.11.13). 

The fourth commit `Update Scylla Java Driver version` and deals with two breaking API
changes of the driver.

PR #299, opened 2022-06-20
Sync with Cassandra 3.11.13 and update a few dependencies (v2)

avelanarius:
This is v2 of #297 as #297 caused test issues after merging.

The first commit `Sync with Cassandra 3.11.13` merges the cassandra-3.11 branch 
from upstream Cassandra repository at the last commit of Cassandra 3.11.13. It also deals 
with the only "real conflict" due to the merge: new method `maybeWarnOnOptions()` had
to be implemented in one of our own classes (`EverywhereStrategy`).

The second commit: `Remove lib/ directory and its JAR files` deals with the fact that
after 2fdf1d27620fe2bce2a3c5989beab40b05fda1e6 appeared in our repository,
dependency management is no longer handled by putting JAR blobs to `lib/`. Instead,
it's fetched dynamically from Maven and put in that folder at the build time. The commit
from upstream repository did remove this directory, but this commit was only partially merged
before to our fork.

Having this folder present in the repo means that some older 
dependencies will linger around - new dependency handling will just
add new JAR files (newer version of dependenecy) and leave the old one
unchanged. This could cause warnings of security scanners.

The third commit `Update version of Netty dependency`. This commit was only present
on Cassandra 4.x branch (not 3.11.13). 

The fourth commit `Update Scylla Java Driver version` and deals with three breaking API
changes of the driver.


avelanarius:
changes in v2:
- copying of `lib/licenses` removed from `dist/debian/debian/scylla-tools-core.install`, `dist/redhat/scylla-tools.spec` and `install.sh`
- cassandra-stress: changed the default protocol version from `NEWEST_SUPPORTED` to `DEFAULT`. `NEWEST_SUPPORTED` in Scylla Java Driver is V5, while `DEFAULT` is V4. The old default of `NEWEST_SUPPORTED` did not work in Scylla, as Scylla does not support V5. In Scylla Java Driver we had two choices a couple months ago: either set `NEWEST_SUPPORTED` to V4 (even though driver supports V5) or keep it at `V5` and introduce new `DEFAULT` constant with V4. Both choices have some disadvantages, but we (with @haaawk) decided on keeping `NEWEST_SUPPORTED` to V5. So far we had no complaints from users about this change, but some uses of the driver that hardcode `NEWEST_SUPPORTED` have to change it to `DEFAULT` (or better: do not set it explicitly!).

The Scylla CI tests now pass correctly with my changes: https://github.com/scylladb/scylla/pull/10768 (a couple of first tries failed due to unrelated flakiness of master).

PR #300, opened 2022-06-27
build_reloc.sh: rename relocatable packages

syuu1228:
Currently, we use following naming convention for relocatable package
filename:
  ${package_name}-${arch}-package-${version}.${release}.tar.gz
But this is very different with Linux standard packaging system such as
.rpm and .deb.
Let's align the convention to .rpm style, so new convention should be:
  ${package_name}-${version}-${release}.${arch}.tar.gz


syuu1228:
@avikivity ping, we need this for https://github.com/scylladb/scylla/pull/10891

Issue #87, opened 2019-03-04
"six" python library in lib is out-of-date with its usage.


elcallio:
The "six" python 2/3 compatibility lib that is included in the "lib" directory is in fact out-of-date with the usage of it in the cassandra python driver. Origin probably never noticed this, because they never ran the affected code paths. 
We, however, do. 

Fix is simple update.

PR #302, opened 2022-07-27
scylla-tools-java: Update "six" library used by cqlsh/python driver

fruch:
commit https://github.com/scylladb/scylla-tools-java/commit/2fdf1d27620fe2bce2a3c5989beab40b05fda1e6 from upstream
remove the usage of libraries in ./lib, and fetching them using
`build-resolver.xml`

commit https://github.com/scylladb/scylla-tools-java/commit/8335869e2de98c7167f81a66cef734914e1ec3e4 downloaded newer
version of six to `./lib`, but now need to update the `build-resolver.xml`
with the correct version we need

Fixes: https://github.com/scylladb/scylla-tools-java/issues/87


avikivity:
Does it fix a bug?

fruch:
> Does it fix a bug?

it's fixes https://github.com/scylladb/scylla-enterprise/issues/731 again, but in a slightly different place

avikivity:
Please update the changelog to explain what it fixes (without referring to an enterprise issue that no one can see).

If you think it will require backporting, please create an issue in scylla.git.

fruch:
> Please update the changelog to explain what it fixes (without referring to an enterprise issue that no one can see).
> 
> If you think it will require backporting, please create an issue in scylla.git.

as far as I know it doesn't need backporting

PR #306, opened 2022-09-04
dist/debian:add python3 as dependency

yaronkaikov:
In https://github.com/scylladb/scylladb/pull/11440 we will start building Scylla docker images based on Ubuntu:22.04, adding `python3` as a dependency since `python` package is not available anymore

### DIFF
--- a/dist/debian/control.template
+++ b/dist/debian/control.template
@@ -17,6 +17,6 @@ Description: Scylla database tools
 
 Package: %{product}-tools-core
 Architecture: all
-Depends: openjdk-8-jre-headless | openjdk-8-jre | oracle-java8-set-default | adoptopenjdk-8-hotspot-jre | openjdk-11-jre-headless | openjdk-11-jre | oracle-java11-set-default, python (>=2.7) | python2, procps
+Depends: openjdk-8-jre-headless | openjdk-8-jre | oracle-java8-set-default | adoptopenjdk-8-hotspot-jre | openjdk-11-jre-headless | openjdk-11-jre | oracle-java11-set-default, python (>=2.7) | python2 | python3, procps
 Description: Scylla database tools core files
  Core files for scylla database tools


nyh (approved):
The patch looks good, but does Python3 actually work for the current Scylla tools, especially cqlsh?

yaronkaikov:
> The patch looks good, but does Python3 actually work for the current Scylla tools, especially cqlsh?

it looks like `cqlsh` require `python2.7` to be installed. updated https://github.com/scylladb/scylladb/pull/11440

nyh:
> > The patch looks good, but does Python3 actually work for the current Scylla tools, especially cqlsh?
> 
> it looks like `cqlsh` require `python2.7` to be installed. updated [scylladb/scylladb#11440](https://github.com/scylladb/scylladb/pull/11440)

So I don't understand - is **this** PR still correct?

yaronkaikov:
> > > The patch looks good, but does Python3 actually work for the current Scylla tools, especially cqlsh?
> > 
> > 
> > it looks like `cqlsh` require `python2.7` to be installed. updated [scylladb/scylladb#11440](https://github.com/scylladb/scylladb/pull/11440)
> 
> So I don't understand - is **this** PR still correct?

yes, it is.

nyh:
But this change suggests that "python3" is enough, while it isn't: cqlsh, part of scylla-tools-java - can't (currently...) run on python3.
Or maybe the python thing is only used to **build** scylla-tools-java, and doesn't say anything about what is required to **run** it?

yaronkaikov:
> But this change suggests that "python3" is enough, while it isn't: cqlsh, part of scylla-tools-java - can't (currently...) run on python3. Or maybe the python thing is only used to **build** scylla-tools-java, and doesn't say anything about what is required to **run** it?

@nyh you are correct , it's only related to the **build** section

fruch:
@yaronkaikov @nyh this need to be yanked out, cqlsh doesn't support python3, and cause of this change our AMIs now don't have python2, and cqlsh fails !!!!

fruch:
> @yaronkaikov @nyh this need to be yanked out, cqlsh doesn't support python3, and cause of this change our AMIs now don't have python2, and cqlsh fails !!!!

FYI, see https://github.com/scylladb/scylladb/issues/11435 for more place we notice there's no more python2.

but we can't break it like that, not until we'll have the replacement for the old cqlsh code in place.

nyh:
I mentioned this above, but @yaronkaikov "promised" me above that it only affects the **build**, not the runtime in the AMI.

The patch is already in so please submit a new patch to fix it. I'm not sure if the new patch should be to scylla-tools-java (adding python2 back), or to the AMI (which needs python2 to run stuff - not to build it).

nyh:
> FYI, see [scylladb/scylladb#11435](https://github.com/scylladb/scylladb/issues/11435) for more place we notice there's no more python2.

Isn't this a separate (and somewhat opposite) problem - i.e., it's a system where Python2 isn't available at all, so we can't ask to install Python2 even if we wanted to?

fruch:
> > FYI, see [scylladb/scylladb#11435](https://github.com/scylladb/scylladb/issues/11435) for more place we notice there's no more python2.
> 
> Isn't this a separate (and somewhat opposite) problem - i.e., it's a system where Python2 isn't available at all, so we can't ask to install Python2 even if we wanted to?

it's all boils to the same issue, our cqlsh code isn't ready to be used with python3, if a distro version doesn't have python2 anymore, we first need to address the issue with cqlsh, and only after that we can move to those distro versions.

anyhow the OR here isn't correct, the tools doesn't need python3 (at least not right now), we have a relocatable python3 that we probably should be using once we cqlsh work for python3.

I think that meanwhile the whole move to ubuntu 22.04, need to be retracted, until we have this sorted out.

yaronkaikov:
> I mentioned this above, but @yaronkaikov "promised" me above that it only affects the **build**, not the runtime in the AMI.
> 
> The patch is already in so please submit a new patch to fix it. I'm not sure if the new patch should be to scylla-tools-java (adding python2 back), or to the AMI (which needs python2 to run stuff - not to build it).

@nyh My mistake :-1: , please revert this patch

nyh:
@yaronkaikov it seems like everyone wants me to do it, so I can, but this is not the proper process: A "git revert" isn't a dequeue - it's a real patch (undoing the previous one), and it should come with a real commit message explaining **why** the revert was done, and review from people who know what the heck this patch means. I thought the original patch was wrong, but I didn't know enough to resist it, and unfortunately I didn't. I would have liked someone who does know enough about the context of this patch to review this revert.

BUT, all that being said, this time I'll revert it myself. Next time please consider just doing reverts as a normal PR.

yaronkaikov:
> @yaronkaikov it seems like everyone wants me to do it, so I can, but this is not the proper process: A "git revert" isn't a dequeue - it's a real patch (undoing the previous one), and it should come with a real commit message explaining **why** the revert was done, and review from people who know what the heck this patch means. I thought the original patch was wrong, but I didn't know enough to resist it, and unfortunately I didn't. I would have liked someone who does know enough about the context of this patch to review this revert.
> 
> BUT, all that being said, this time I'll revert it myself. Next time please consider just doing reverts as a normal PR.

@nyh Thanks, and sorry for the mess, next time i will just do the revert

nyh:
Reverted.

PR #307, opened 2022-09-13
pylib/: switch to using scylla-cqlsh

fruch:
python2 package isn't supported anymore in CentOS 9 Stream (python2.7 is EOL for more then 2 years by now)
and newer distributions would also start dropping it
 
We are still depended on python2, cause we are using old cqlsh code from cassandra 3.11 branch.

So we'll fork the cqlsh code out of cassandra/trunk where there a version that already supports python3

* our cqlsh fork https://github.com/scylladb/scylla-cqlsh (temporary it's in https://github.com/fruch/scylla-cqlsh)
* packages would be uploaded to pypi - https://pypi.org/project/scylla-cqlsh/
* `scylla-tools-java` - would still bundle cqlsh for distribution as relocatable/deb/rpm

### changes to scylla-tools-java:
* remove the local cqlsh code
* download scylla-driver,six,pyyaml as source distributions (zip them for import)
* download cqlsh in pylib

Ref: https://github.com/scylladb/scylladb/issues/11435


nyh:
Hi @fruch. You commit message explained what you did, but not **why** you're doing it.

Specifically, why does it make sense to "remove the local cqlsh code" and then have it "downloaded" from somewhere else? Where does the code in that somewhere else come from? Why can't we just move that other place to here, inside scylla-tools.java, or, if we already have an independent git module for this scylla-cqlsh, why don't we make it a submodule (better a parallel submodule in tools/ of scylla.git, but can also be a nested submodule inside scylla-tools-java.git).

Another thing you didn't write is about the immediate and/or long-term implications of this patch of changing the cqlsh. How is the "scylla-cqlsh" different after this patch from the "cqlsh" we had before this patch? Is it a newer version of cqlsh? Does it use Python3 vs. Python2? Or what?

fruch:
> Hi @fruch. Your commit message explained what you did, but not **why** you're doing it.
> 
> Specifically, why does it make sense to "remove the local cqlsh code" and then have it "downloaded" from somewhere else? Where does the code in that somewhere else come from? Why can't we just move that other place to here, inside scylla-tools.java, or, if we already have an independent git module for this scylla-cqlsh, why don't we make it a submodule (better a parallel submodule in tools/ of scylla.git, but can also be a nested submodule inside scylla-tools-java.git).
> 
> Another thing you didn't write is about the immediate and/or long-term implications of this patch of changing the cqlsh. How is the "scylla-cqlsh" different after this patch from the "cqlsh" we had before this patch? Is it a newer version of cqlsh? Does it use Python3 vs. Python2? Or what?

update the commit/PR, with more of the details.

as for sub-module, that's also an option, once @avikivity would move it the organization 

this initial PR was to flush out all the packaging related worked need, now switching it to sub-module for example should be hard.

the extra perks we'll get: 
* we can `pip install scylla-cqlsh` where we need it for testing, even when we don't fulling install scylla packages.
* CI with running the unittest of cqlsh (which we didn't ever used/run)

.build/build-resolver.xml
```diff
@@ -178,12 +178,12 @@
...
         <!-- python-driver -->
-        <get src="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/cassandra-driver-internal-only-3.11.0-bb96859b.zip" dest="${local.repository}/org/apache/cassandra/deps/cassandra-driver-internal-only-3.11.0-bb96859b.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
-
+        <get src="https://files.pythonhosted.org/packages/ae/05/7cb27e35fc7c76fc2df845eb7509e540238067cc34888d57c93d787b2fe7/scylla-driver-3.25.5.tar.gz" dest="${local.repository}/org/apache/cassandra/deps/scylla-driver-3.25.5.tar.gz" usetimestamp="true" quiet="true" skipexisting="true"/>
+        <!-- scylla-cqlsh -->
+        <get src="https://files.pythonhosted.org/packages/76/6e/23f2b1d41b5acb0c5745d2fed6d3f2dc14ebce952d3d6e53787ee1b9da93/scylla-cqlsh-6.0.2.tar.gz" dest="${local.repository}/org/apache/cassandra/deps/scylla-cqlsh-6.0.2.tar.gz" usetimestamp="true" quiet="true" skipexisting="true"/>
```
avikivity, line 186: This is unmaintainable. It should be a submodule so there is a single hash to maintain.

Why does it need to be under scylla-tools-java.git? We should move it out to scylladb.git.

avikivity: Though if it's easier, we could do it in two steps, first under scylla-tools-java.git, then moving it out to scylladb.git (and create new subpackages scylla-cqlsh.{rpm.deb}.

nyh:
> this initial PR was to flush out all the packaging related worked need

Just a few days ago I committed a patch that sounded "reasonable" but its intended (and unintended!) effects were really in other super-packages that used that one, and it turned out the effects were bad, not good. This is why I very hesitate to commit this patch without understanding its intended - and unintended, effects.

> the extra perks we'll get:
> 
>     * we can `pip install scylla-cqlsh` where we need it for testing, even when we don't fulling install scylla packages.

But that's completely unrelated to this patch... Wherever the scylla-cqlsh source code comes from, one of us could contribute it also to pypi, and indeed other packages can get it from there and not from source code.

But I think the most important thing to consider is where our **source code** is, and we need to have a good process for modifying this source code - not just using it. I don't think it's a good practice that our source-code repository (which scylla-tools-java is) uses other source code as some "magic things that comes from who-knows-where".

I'm trying to understand if there's a reason why not to move scylla-cqlsh **into** scylla-tools-java like it always was, not even as a submodule? Yes, the name "scylla-tools-java" is a misnomer, but it was also a misnomer in the past.

avikivity:
> > this initial PR was to flush out all the packaging related worked need
> 
> Just a few days ago I committed a patch that sounded "reasonable" but its intended (and unintended!) effects were really in other super-packages that used that one, and it turned out the effects were bad, not good. This is why I very hesitate to commit this patch without understanding its intended - and unintended, effects.
> 
> > the extra perks we'll get:
> > ```
> > * we can `pip install scylla-cqlsh` where we need it for testing, even when we don't fulling install scylla packages.
> > ```
> 
> But that's completely unrelated to this patch... Wherever the scylla-cqlsh source code comes from, one of us could contribute it also to pypi, and indeed other packages can get it from there and not from source code.
> 
> But I think the most important thing to consider is where our **source code** is, and we need to have a good process for modifying this source code - not just using it. I don't think it's a good practice that our source-code repository (which scylla-tools-java is) uses other source code as some "magic things that comes from who-knows-where".

We should certainly have our build process use the sources directly, not sources -> offline build -> upload somewhere -> download -> use.

> 
> I'm trying to understand if there's a reason why not to move scylla-cqlsh **into** scylla-tools-java like it always was, not even as a submodule? Yes, the name "scylla-tools-java" is a misnomer, but it was also a misnomer in the past.

scylla-tools-java is a fork of Cassandra that we use for tools. If we start to develop something independently, it should be in a separate repository to avoid carrying over all that unused code and so that we can eventually get rid of scylla-tools-java.

nyh:
>     * our cqlsh fork https://github.com/scylladb/scylla-cqlsh (temporary it's in https://github.com/fruch/scylla-cqlsh)

I suggest that this should be the first thing that we fix - before this PR.
Or not have a new repository at all - and just a subdirectory in the misnamed scylla-tools-java repository (as we had today).

I think we should not replace a pre-existing **official** part of our distribution by something coming from a non-official source.

>     * packages would be uploaded to pypi - https://pypi.org/project/scylla-cqlsh/

The second thing I think we should do before this PR is to make **this** process official as well. How does this project get built and sent to pypi - by whom or by which process? How is this process reliable or secure?

Or better yet, because of the security and/or process implications, I think we shouldn't do it at all. I mean, sure, we can (and should!) send scylla-cqlsh to pypi, but we shouldn't rely on this in our own builds. It's silly that Scylla needs to write its code into pypi and then read it back just to build its own package. I think our official builds should take scylla-cqlsh from our own source code repository (as submodule or whatever), not from pypi.

Of course, other pieces of our build, such as tests, may get the code from pypi. But I don't think our official package builds should do that.

>     * `scylla-tools-java` - would still bundle cqlsh for distribution as relocatable/deb/rpm

I wonder why. If we move cqlsh to a separate repository, why should pieces of its package
building still remain in scylla-tools-java? Is it just some temporary workaround you want to
fix later, or the way you plan to leave it?

.build/build-resolver.xml
fruch: I'll would move it to a submodule, that's should be easy part

bundling cqlsh it into it's own artifact probably going to have a ripple effect
* for example ccm won't be aware of and cqlsh would be missing

avikivity: It would still be part of scylla.tar.gz and scylla.rpm would pull in scylla-cqlsh.rpm, so no impact to users.

fruch: @avikivity, I've started with transfering https://github.com/fruch/scylla-cqlsh to you, so you can add it to scylladb.

fruch:
> The second thing I think we should do before this PR is to make **this** process official as well. How does this project get built and sent to pypi - by whom or by which process? How is this process reliable or secure?

it would be the same process we are building python-driver

https://github.com/scylladb/python-driver/blob/master/.github/workflows/build-push.yml#L139

with the official github actions for that, building only tags create on the master branch.
so only whom have access to the master branch to tag it, can initiate a release.

.build/build-resolver.xml
fruch: > It would still be part of scylla.tar.gz and scylla.rpm would pull in scylla-cqlsh.rpm, so no impact to users.

if everything goes smoothly maybe, 
as for the tar.gz... ccm doesn't use the unified one, at least not yet. it would need changes.

so with that the number of repos that would need to get changed and adapt is 
* scylla-tools-java - pure anything cqlsh
* scylla-cqlsh - create packaging for deb/rpm/tar.gz
* scylla-ccm - move to use unified package, and install all of the packages
* scylla - add the needed ninja target 
* scylla-pkg - verify new artifact are known an collected

avikivity:
If it has its own well defined release process then maybe we can just pick the binaries (like in the patch). I'm not comfortable with it, but we don't (and can't) pull in all the dependencies.

fruch:
> If it has its own well defined release process then maybe we can just pick the binaries (like in the patch). I'm not comfortable with it, but we don't (and can't) pull in all the dependencies.

we need to pull in the dependencies, like the python-driver (scylla-driver in our case), since it doesn't have RPMs/DEBs, and we need it as source package, since can be compiled only against  specific python versions, i.e. some part won't work if we'll build for python 3.10, but try to run it with python3.8.

the rest of the dependencies are also for when it's use as relocatable package, when you don't have dependencies

I'm not sure I understood "but we don't (and can't) pull in all the dependencies."...

fruch:
closing in favor of https://github.com/scylladb/scylladb/pull/11937

Issue #310, opened 2022-09-19
install.sh is using wrong permissions for install cqlsh files


fruch:
`install.sh` is using copy, and not install
```
cp -r pylib/cqlshlib "$rprefix"/share/cassandra/pylib
```

on oel81 when installing unified relocatable package, cqlsh is failing to run from a user which isn't root (when root did the install)
```
07:45:21  Command: 'cqlsh --no-color   --request-timeout=120 --connect-timeout=60  -e "desc keyspaces" 10.4.2.93 9042'
07:45:21  Exit code: 1
07:45:21  Stdout:
07:45:21  Stderr:
07:45:21  /etc/bashrc: line 99: TMOUT: readonly variable
07:45:21  /etc/bashrc: line 100: TMOUT: readonly variable
07:45:21  Traceback (most recent call last):
07:45:21  File "/opt/scylladb/share/cassandra/bin/cqlsh.py", line 169, in <module>
07:45:21  from cqlshlib import cql3handling, cqlhandling, pylexotron, sslhandling, cqlshhandling
07:45:21  ImportError: No module named cqlshlib
```

the command that should be used:
```
find pylib/cqlshlib -type f -exec install -Dm 755 "{}" "$rprefix"/share/cassandra/{} \;
```

fruch:
example of such a failure:
https://jenkins.scylladb.com/job/scylla-master/job/artifacts-offline-install/job/artifacts-oel81-test/57/

PR #311, opened 2022-09-19
install.sh is using wrong permissions for install cqlsh files

yaronkaikov:
`install.sh` is using copy, and not install
```
cp -r pylib/cqlshlib "$rprefix"/share/cassandra/pylib
```

on oel81 when installing unified relocatable package, cqlsh is failing to run from a user which isn't root (when root did the install)
```
07:45:21  Command: 'cqlsh --no-color   --request-timeout=120 --connect-timeout=60  -e "desc keyspaces" 10.4.2.93 9042'
07:45:21  Exit code: 1
07:45:21  Stdout:
07:45:21  Stderr:
07:45:21  /etc/bashrc: line 99: TMOUT: readonly variable
07:45:21  /etc/bashrc: line 100: TMOUT: readonly variable
07:45:21  Traceback (most recent call last):
07:45:21  File "/opt/scylladb/share/cassandra/bin/cqlsh.py", line 169, in <module>
07:45:21  from cqlshlib import cql3handling, cqlhandling, pylexotron, sslhandling, cqlshhandling
07:45:21  ImportError: No module named cqlshlib
```
reported by @fruch 

Closes: https://github.com/scylladb/scylla-tools-java/issues/310

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -122,7 +122,7 @@ fi
 install -d -m755 "$rprefix"/share/cassandra/pylib
 install -d -m755 "$retc"/bash_completion.d
 install -m644 dist/common/nodetool-completion "$retc"/bash_completion.d
-cp -r pylib/cqlshlib "$rprefix"/share/cassandra/pylib
+cp -rp pylib/cqlshlib "$rprefix"/share/cassandra/pylib
 for i in bin/{nodetool,sstableloader,cqlsh,cqlsh.py,scylla-sstableloader} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
     bn=$(basename $i)
     install -m755 $i "$rprefix"/share/cassandra/bin


install.sh
```diff
@@ -122,7 +122,7 @@ fi
 install -d -m755 "$rprefix"/share/cassandra/pylib
 install -d -m755 "$retc"/bash_completion.d
 install -m644 dist/common/nodetool-completion "$retc"/bash_completion.d
-cp -r pylib/cqlshlib "$rprefix"/share/cassandra/pylib
+find pylib/cqlshlib -type f -exec install -Dm 755 "{}" "$rprefix"/share/cassandra/{} \;
```
nyh, line 125: Is mode 755 the appropriate mode for all files? Also, isn't mode 755 (rwxr-xr-x) the default anyway for "install"?

I wonder if adding "-p" to the cp command (to copy the mode from the source file) wouldn't have solved the permission problem more easily. But "install" is fine too.

yaronkaikov: @nyh you were right about the `-p` addition. 
Verified with https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/artifacts-offline-install/job/artifacts-oel81-test/30/

nyh (approved):

PR #314, opened 2022-11-03
cassandra-stress: pass all hosts straight to the driver

fruch:
In cases of multiple contact points configued from commandline code was randomly selecting one host and pass that to the cql driver if for some reason that specfic node wasn't available the whole command would fail.

removing this random, and passing all hosts to the driver fix this issue, and the driver code is more then capble of handling a situation of some contact points unreachable

### DIFF
--- a/tools/stress/src/org/apache/cassandra/stress/settings/StressSettings.java
+++ b/tools/stress/src/org/apache/cassandra/stress/settings/StressSettings.java
@@ -205,12 +205,11 @@ public class StressSettings implements Serializable
 
             try
             {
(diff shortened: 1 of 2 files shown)


Bouncheck (approved):
LGTM
Ran locally quick write workload with a list of a few unreachable hosts and 1 reachable. Connects just fine.

fruch:
> LGTM Ran locally quick write workload with a list of a few unreachable hosts and 1 reachable. Connects just fine.

yep, exactly how I checked it too.

Issue #318, opened 2022-11-09
system_distributed_everywhere should not be shown in "desc schema"


tarzanek:
"desc schema" doesn't currently show any system* tables
but since 5.0(or earlier? 4.0 definitely didn't have it) it started showing 
```
CREATE KEYSPACE system_distributed_everywhere WITH replication = {'class': 'EverywhereStrategy'}  AND durable_writes = true;

CREATE TABLE system_distributed_everywhere.cdc_generation_descriptions_v2 (
...
```

and I don't think this belongs there.
"desc schema" is used for user tables, system tables have nothing to do there

avelanarius:
The reason of the bug: https://github.com/scylladb/scylla-tools-java/blob/caf754f2434a55ad04e9dd147b8d97c62b8d8427/pylib/cqlshlib/cql3handling.py#L38

`SYSTEM_KEYSPACES` is missing `system_distributed_everywhere`

PR #320, opened 2022-11-25
Mark "system_distributed_everywhere" as system keyspace

avelanarius:
Add `system_distributed_everywhere` keyspace to `SYSTEM_KEYSPACES` list, which marks which keyspaces are recognized as system keyspaces.

This controls for example which keyspaces are not shown when executing `DESC SCHEMA` command in cqlsh. Before this change, `DESC SCHEMA` would print the schema of `system_distributed_everywhere`, which was incorrect as `DESC SCHEMA` is supposed to print only user tables.

Fixes scylladb/scylla-tools-java#318

### DIFF
--- a/pylib/cqlshlib/cql3handling.py
+++ b/pylib/cqlshlib/cql3handling.py
@@ -35,7 +35,7 @@ class UnexpectedTableStructure(UserWarning):
         return 'Unexpected table structure; may not translate correctly to CQL. ' + self.msg
 
 
-SYSTEM_KEYSPACES = ('system', 'system_schema', 'system_traces', 'system_auth', 'system_distributed')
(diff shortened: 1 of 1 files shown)


avikivity:
Please follow up with a similar change in scylladb/scylla-cqlsh.

nyh:
@avikivity how come this PR with just a single patch is recognized by github as "merged" instead of "close"? Did you change something in your merging process?

avikivity:
This is a patch for scylla-tools-java, not scylladb.git. I used the "rebase and merge" to rebase it directly into master (and then queued the submodule update into scylladb.git next).

The problem with our workflow is that patches are targeted to master (so that Fixes # only happens on promotion), but we apply them to next.

Issue #330, opened 2023-04-18
Incorrect sstablesPerRead values


Deexie:
TableHistograms::execute(NodeProbe) sets incorrect percentile values 
in sstablesPerRead array, causing that nodetool's tablehistogram
shows deceiving SSTables values.

bhalevy:
@scylladb/scylla-maint please backport to all live branches

bhalevy:
Cc @DoronArazii

PR #331, opened 2023-04-18
Use EstimatedHistogram in metricPercentilesAsArray(JmxHistogramMBean)

Deexie:
NodeProbe::metricPercentilesAsArray(JmxHistogramMBean) is expected
to return an array with percentiles obtained from a given histogram.
Currently BufferSamples object is created out of the original jmx
histogram's values and its getValue method is used to create an array.
BufferSamples::getValue(double) expects data in different format than
the provided one, though.

EstimatedHistogram, which is compactible with original histogram,
is created out of metric values.

Fixes: #330.

### DIFF
--- a/src/java/org/apache/cassandra/tools/NodeProbe.java
+++ b/src/java/org/apache/cassandra/tools/NodeProbe.java
@@ -101,6 +101,7 @@ import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.Uninterruptibles;
 import org.apache.cassandra.tools.nodetool.GetTimeout;
+import org.apache.cassandra.utils.EstimatedHistogram;
 
 /**
  * JMX client operations for Cassandra.
@@ -1594,15 +1595,15 @@ public class NodeProbe implements AutoCloseable
 
     public double[] metricPercentilesAsArray(CassandraMetricsRegistry.JmxHistogramMBean metric)
     {
-        BufferSamples bs = new BufferSamples(metric.values());
+        EstimatedHistogram eh = new EstimatedHistogram(metric.values());
 
-        return new double[]{ bs.getValue(0.5),
-                bs.getValue(0.75),
-                bs.getValue(0.95),
-                bs.getValue(0.98),
-                bs.getValue(0.99),
-                bs.getMin(),
-                bs.getMax()};
+        return new double[]{ eh.percentile(0.5),
+                eh.percentile(0.75),
+                eh.percentile(0.95),
+                eh.percentile(0.98),
+                eh.percentile(0.99),
+                eh.min(),
+                eh.max()};
     }
 
     public double[] metricPercentilesAsArray(CassandraMetricsRegistry.JmxTimerMBean metric)


Deexie:
@tarzanek @amnonh @bhalevy review please

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -1594,15 +1594,13 @@ public long getStorageMetric(String metricName)
 
     public double[] metricPercentilesAsArray(CassandraMetricsRegistry.JmxHistogramMBean metric)
     {
-        BufferSamples bs = new BufferSamples(metric.values());
```
tarzanek, line 1597: I expect this was to get some caching done, by removing this we will loose this caching
for direct fetching (ala histograms) it won't be a prob
but I can imagine this can be a prob e.g. for e.g. toppartitions which does sampling and polls periodically, can you test your change with toppartitions? ( or if it doesn't use this call, then check all possible calls and make sure removing caching won't break them)

tarzanek: ref https://github.com/scylladb/scylla-tools-java/commit/ee5e2b970dadda951ffc3f375f917bd0d8af49ca
this explains why, so you are reverting Amnons optimalization ;-)

Deexie: Oh, I see. How about moving the logic to jmx side? I mean adding `metricPercentilesAsArray` to jmx `JmxHistogramMBean` interface and then 
```
public double[] metricPercentilesAsArray(CassandraMetricsRegistry.JmxHistogramMBean metric) {
    return metric.metricPercentilesAsArray();
}
```
We will have one call to jmx and be able to differentiate between the versions of histogram.

amnonh:
Inside Scylla, we have the old-style histograms we inherit from Cassandra and the new kind (actual histograms) these days for most of them we even have the derivative histogram, which is what we are really after in this case.
Maybe this is the path we should take?

Move to use the real histograms and deprecate the old kind.

bhalevy:
@Deexie and I reviewed the code again today and it looks the bug is the intrrpretation of the JMXHistogramMBean values on the Nodetool NodeProbe side as buffer samples.

Let's see if we can reconstruct the EstimatedHistogram from values() using the HistogramBuilder and use it to retrieve the percentiles.

This way we preserve the fix of getting the values in a single call rather than a series of calls to get each percentile, and we do the conversion correctly.

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -1594,15 +1595,15 @@ public long getStorageMetric(String metricName)
...
-                bs.getValue(0.99),
-                bs.getMin(),
-                bs.getMax()};
+        return new double[]{ eh.percentile(0.5),
+                eh.percentile(0.75),
+                eh.percentile(0.95),
+                eh.percentile(0.98),
+                eh.percentile(0.99),
+                eh.min(),
+                eh.max()};
```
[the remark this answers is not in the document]
Deexie: Yes, it does :D The results look reasonable and the test passes

bhalevy (approved):

PR #340, opened 2023-09-05
install-dependencies.sh: do not install weak dependencies

mykaul:
Especially for Java, we really do not need the tens of packages and MBs it adds, just because Java apps can be built and use sound and graphics and whatnot.


denesb (approved):
LGTM

I think this will require a frozen toolchain update, so I'll leave merging it to @avikivity.

mykaul:
> LGTM
> 
> I think this will require a frozen toolchain update, so I'll leave merging it to @avikivity.

Correct, but that's the next step after this gets in, we can (should) update the toolchain (same cycle as https://github.com/scylladb/scylladb/pull/15045 ?)

avikivity:
No, it has to be done atomically (well, the sky won't fall down if we don't, but it's the Right Thing To Do).

avikivity:
https://github.com/scylladb/scylladb/pull/15322

PR #342, opened 2023-09-13
nodetool: parse and forward -h|--host to nodetool

denesb:
Our launcher script (bin/nodetool) currently silently ignores the host parameter and nodetool will use the default value (127.0.0.1). This patch parses and forwards the host parameter to nodetool, so that it is possible to start JMX on other ips than 127.0.0.1.


bhalevy (approved):
LGTM

bhalevy:
@scylladb/scylla-tools-maint please review/merge

denesb:
@scylladb/scylla-tools-maint ping.

PR #344, opened 2023-09-26
resolver: download sigar artifact only for Linux / AMD64

mykaul:
We don't need or support the other platforms.
This removes multi MBs and multiple downloads from Maven, that are not needed during build.

### DIFF
--- a/.build/build-resolver.xml
+++ b/.build/build-resolver.xml
@@ -179,53 +179,11 @@
 
         <!-- apache/cassandra/lib -->
         <get dest="${local.repository}/org/apache/cassandra/deps/sigar-bin/" quiet="true" usetimestamp="true" skipexisting="true">
-            <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-amd64-freebsd-6.so"/>
             <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-amd64-linux.so"/>
-            <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-amd64-solaris.so"/>
-            <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-ia64-hpux-11.sl"/>
-            <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-ia64-linux.so"/>
-            <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-pa-hpux-11.sl"/>
-            <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-ppc-aix-5.so"/>
-            <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-ppc-linux.so"/>
-            <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-ppc64-aix-5.so"/>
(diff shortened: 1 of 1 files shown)


.build/build-resolver.xml
```diff
@@ -179,53 +179,11 @@
 
         <!-- apache/cassandra/lib -->
         <get dest="${local.repository}/org/apache/cassandra/deps/sigar-bin/" quiet="true" usetimestamp="true" skipexisting="true">
-            <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-amd64-freebsd-6.so"/>
             <url url="https://raw.githubusercontent.com/apache/cassandra/${lib.download.sha}/lib/sigar-bin/libsigar-amd64-linux.so"/>
```
avikivity, line 182: We support ARM too. Is this download optional? Can we skip it for x86 too?

mykaul: I only removed deps, not added ones. 
I could not find one in Cassandra's either (https://github.com/apache/cassandra/blob/cfe48de0196993d13d25a799c4d13ac3309d5ae6/.build/build-resolver.xml#L242C81-L242C108 )

avikivity: Maybe we can drop the whole thing. 

```java
    private SigarLibrary()
    {
        logger.info("Initializing SIGAR library");
        try
        {
            sigar = new Sigar();
            mounts = sigar.getFileSystemMap();
            initialized = true;
        }
        catch (SigarException e)
        {
            logger.info("Could not initialize SIGAR library {} ", e.getMessage());
        }
        catch (UnsatisfiedLinkError linkError)
        {
            logger.info("Could not initialize SIGAR library {} ", linkError.getMessage());
        }
    }
```

So it will just log something and continue. In any case we don't need that since we don't run Cassandra itself, just cassandra-stress and nodetool.

mykaul: I can just remove the whole code:
```
[ykaul@ykaul scylla-tools-java]$ grep -Frisn SigarLibrary
test/distributed/org/apache/cassandra/distributed/test/ResourceLeakTest.java:42:import org.apache.cassandra.utils.SigarLibrary;
test/distributed/org/apache/cassandra/distributed/test/ResourceLeakTest.java:89:        long pid = SigarLibrary.instance.getPid();
src/java/org/apache/cassandra/utils/SigarLibrary.java:24:public class SigarLibrary
src/java/org/apache/cassandra/utils/SigarLibrary.java:26:    private Logger logger = LoggerFactory.getLogger(SigarLibrary.class);
src/java/org/apache/cassandra/utils/SigarLibrary.java:28:    public static final SigarLibrary instance = new SigarLibrary();
src/java/org/apache/cassandra/utils/SigarLibrary.java:42:    private SigarLibrary()
src/java/org/apache/cassandra/service/StartupChecks.java:55:import org.apache.cassandra.utils.SigarLibrary;
src/java/org/apache/cassandra/service/StartupChecks.java:64: * (initSigarLibrary, checkCacheServiceInitialization) and invariant checking
src/java/org/apache/cassandra/service/StartupChecks.java:92:                                                                      initSigarLibrary,
src/java/org/apache/cassandra/service/StartupChecks.java:262:    public static final StartupCheck initSigarLibrary = new StartupCheck()
src/java/org/apache/cassandra/service/StartupChecks.java:266:            SigarLibrary.instance.warnIfRunningInDegradedMode();
```

But I think we don't wish to test the Java code right now - or do we ?

avikivity: This is used in Cassandra itself, not in nodetool. We'll never run this code.

mykaul:
@avikivity - is my change reasonable enough, and in a followup will get rid of Sigar completely?

avikivity:
Ok

.build/build-resolver.xml
mykaul: Will it introduce merge challenges, if we ever merge from upstream? Otherwise, we can remove most of Cassandra as well.

mykaul:
@avikivity - I think this can be merged as is, then we can remove sigar altogether, which is slightly more work and risk (due to lack of tests all around)

PR #345, opened 2023-10-01
Fix compilation with Java 11

mykaul:
This is an effort to compile using Java 11 properly - set JAVA11_HOME env. variable and use it in the build XML - just as was done with Java 8.

* needs to be tested via a CI build *

### DIFF
--- a/reloc/build_reloc.sh
+++ b/reloc/build_reloc.sh
@@ -77,6 +77,15 @@ if [ -z "$JAVA8_HOME" ]; then
         fi
     done
 fi
+if [ -z "$JAVA11_HOME" ]; then
+    for i in /usr/lib/jvm/java-11
+    do
+        if [ -e "$i" ]; then
+            export JAVA11_HOME="$i"
+            break
+        fi
+    done
+fi
 
 ant jar
 dist/debian/debian_files_gen.py
--- a/build.xml
+++ b/build.xml
@@ -836,10 +836,11 @@
                 <path refid="cassandra.classpath"/>
             </classpath>
         </javac>
-        <echo message="Compiling for current java version ..."/>
+        <echo message="Compiling for Java 11 (using ${env.JAVA11_HOME}/bin/javac) ..."/>
         <javac fork="true"
                debug="true" debuglevel="${debuglevel}" encoding="utf-8"
                destdir="${build.classes.main}/META-INF/versions/11" includeantruntime="false"
+               executable="${env.JAVA11_HOME}/bin/javac"
                memorymaximumsize="512M">
             <src path="${build.src.java11}"/>
             <compilerarg value="--release"/>


mykaul:
@yaronkaikov - what's the proper way to run CI pipeline on it, to build this stuff?

yaronkaikov:
You can apply those changes on your Scylla fork and open a PR so the CI will start running

Or use https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/

mykaul:
> You can apply those changes on your Scylla fork and open a PR so the CI will start running

I'll need to understand better how to do it - as this is a submodule after all...
> 
> Or use https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/

yaronkaikov:
@mykaul I have build your change locally and created a docker image, when running it locally it seems to be working
Also `nodetool status` , not sure if it's enough

```
[yaronkaikov@london]~/git/scylladb (debug-submodule)$ podman ps -a
CONTAINER ID  IMAGE         COMMAND     CREATED         STATUS         PORTS       NAMES
59e43d45f571  5d5a0cfe8a77              27 seconds ago  Up 27 seconds              dreamy_lamport
[yaronkaikov@london]~/git/scylladb (debug-submodule)$ podman exec -it dreamy_lamport /bin/bash

   _____            _ _       _____  ____  
  / ____|          | | |     |  __ \|  _ \ 
 | (___   ___ _   _| | | __ _| |  | | |_) |
  \___ \ / __| | | | | |/ _` | |  | |  _ < 
  ____) | (__| |_| | | | (_| | |__| | |_) |
 |_____/ \___|\__, |_|_|\__,_|_____/|____/ 
               __/ |                       
              |___/                        


Nodetool:
	nodetool help
CQL Shell:
	cqlsh
More documentation available at: 
	http://www.scylladb.com/doc/

root@59e43d45f571:/# nodetool status
Datacenter: datacenter1
=======================
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address     Load       Tokens       Owns    Host ID                               Rack
UN  10.0.2.100  162.22 KB  256          ?       d94a4a10-1f70-4b87-aec1-889394e756ea  rack1

Note: Non-system keyspaces don't have the same replication settings, effective ownership information is meaningless
root@59e43d45f571:/# 
```

mykaul:
Evidently it failed building in your test PR, so I did not get something right. It needs to compile it well with Java 11. I'm unsure why we still need Java 8, if anything.

mykaul:
Could be that it was not built right and is still using Java 8 build?

avikivity:
We need to drop the version dependency. It's there in Cassandra because it was using undocumented internals which are version dependent. But since we only use it for nodetool, we can just comment out these parts.

We'll still need to target some baseline JVM version, but 11 seems to be old enough.

@elcallio please correct any mistakes here.

yaronkaikov:
> Evidently it failed building in your test PR, so I did not get something right. It needs to compile it well with Java 11. I'm unsure why we still need Java 8, if anything.

Well i removed the java8 from the docker and it's still working 

```
[yaronkaikov@london]~/git/scylladb (debug-submodule)$ git diff
diff --git a/dist/docker/debian/build_docker.sh b/dist/docker/debian/build_docker.sh
index 28be4044c..f6c87f797 100755
--- a/dist/docker/debian/build_docker.sh
+++ b/dist/docker/debian/build_docker.sh
@@ -77,7 +77,7 @@ run apt-get -y update
 run apt-get -y install dialog apt-utils
 run bash -ec "echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections"
 run bash -ec "rm -rf /etc/rsyslog.conf"
-run apt-get -y install hostname supervisor openssh-server openssh-client openjdk-11-jre-headless python2 python3 python3-yaml curl rsyslog sudo
+run apt-get -y install hostname supervisor openssh-server openssh-client openjdk-11-jre-headless python3 python3-yaml curl rsyslog sudo
 run bash -ec "echo LANG=C.UTF-8 > /etc/default/locale"
 run bash -ec "dpkg -i packages/*.deb"
 run apt-get -y clean all
```

```
[yaronkaikov@london]~/git/scylladb (debug-submodule)$ podman exec -it beautiful_meninsky nodetool status
Datacenter: datacenter1
=======================
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address     Load       Tokens       Owns    Host ID                               Rack
UN  10.0.2.100  163.64 KB  256          ?       f281e23d-c0b0-495f-9043-d6db6b5e18f9  rack1

Note: Non-system keyspaces don't have the same replication settings, effective ownership information is meaningless
```

elcallio:
Can just neuter (remove all java-8-only stuff, effectively disable) AtomicBTreePartitionBase and move to shared source. Then neither specific java8_home or other should be required, just a version check on the jdk used.

avikivity:
@avelanarius please see if your team can handle this

tchaikov:
FWIW, i have a pull request for buiding jmx with JDK-11, see https://github.com/scylladb/scylla-jmx/pull/207

reloc/build_reloc.sh
```diff
@@ -77,6 +77,15 @@ if [ -z "$JAVA8_HOME" ]; then
         fi
     done
 fi
+if [ -z "$JAVA11_HOME" ]; then
+    for i in /usr/lib/jvm/java-11
```
tchaikov, line 81: i am afraid that this does not always work. as on f38, java-11-openjdk-headless does not install into `/usr/lib/jvm/java-11`, it installs its stuff into `/usr/lib/jvm/java-11-openjdk-11.0.20.0.8-1.fc38.x86_64`. the symlink of `/usr/lib/jvm/java-11` is created by `java-11-openjdk-devel`. i think the idea is that this  offers the flexibility to use other variant of jdk-11 implementations via `alternatives`. another option is to use something like https://github.com/scylladb/scylla-jmx/blob/master/scripts/select-java .

mykaul: Isn't the symlink sufficient? And we do (regretfully) install today the openjdk-devel package, no?
https://github.com/scylladb/scylla-jmx/blob/master/scripts/select-java will pick the 1st 'Java 11' version, won't it?
(I think we should use 'alternatives', and not re-invent it)

tchaikov: sorry, i was under the impression that we install `openjdk-11-jdk-headless`. in jmx, we do install the headless package, though. see https://github.com/scylladb/scylla-jmx/blob/770c1f9fb06475fcad2337283e12188b5008a299/install-dependencies.sh#L25

but apparently, this does not apply to this repo where we install `java-1.8.0-openjdk-devel`, which, well, regretfully installs `java-1.8.0-openjdk`. so, we should have the symlink. but if we decided to use the headless package, we won't have the symlink.

> https://github.com/scylladb/scylla-jmx/blob/master/scripts/select-java will pick the 1st 'Java 11' version, won't it?
> (I think we should use 'alternatives', and not re-invent it)

yeah, but we should not install unnecessary bits. if we install the alternative manually, there is chance that it conflict with the one created by  `java-11-openjdk` if user installs this package by himself/herself,  but on the other hand, we don't want to install it, so we should not depend on it.

mykaul: I can't control what the user install by themselves, and I don't think we should support it.

syuu1228:
@mykaul We also need to update install-dependencies.sh like this:
```
diff --git a/install-dependencies.sh b/install-dependencies.sh
index 9272e154b2..b031f5a71b 100755
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -20,7 +20,7 @@
 . /etc/os-release
 
 if [ "$ID" = "ubuntu" ] || [ "$ID" = "debian" ]; then
-    apt -y install openjdk-8-jdk-headless ant ant-optional python
+    apt -y install openjdk-11-jdk-headless ant ant-optional python
 elif [ "$ID" = "fedora" ] || [ "$ID" = "centos" ]; then
-    dnf install -y --setopt=install_weak_deps=False ant java-1.8.0-openjdk-devel python ant-junit fakeroot
+    dnf install -y --setopt=install_weak_deps=False ant java-11-openjdk-devel python ant-junit fakeroot
 fi
```

reloc/build_reloc.sh
```diff
@@ -77,6 +77,15 @@ if [ -z "$JAVA8_HOME" ]; then
...
+    do
+        if [ -e "$i" ]; then
+            export JAVA11_HOME="$i"
```
syuu1228, line 84: I think export JAVA11_HOME should be done in tools/toolchain/Dockerfile of scylla main repo.
Because we currently have `ENV JAVA8_HOME` on the file:
https://github.com/scylladb/scylladb/blob/master/tools/toolchain/Dockerfile#L16

It will automatically apply when we use tools/toolchain/dbuild.

mykaul:
Not sure it's the right thing right now, closing for the time being.

tchaikov:
resurrected at #368

reloc/build_reloc.sh
tchaikov: we still require this change in case user do not use the `dbuild` and use the `build_reloc.sh` directly. and since we need the change in `build_reloc.sh`, i'd suggest to remove `ENV JAVA8_HOME`.

PR #346, opened 2023-10-13
cassandra-stress: delay before retry

tchaikov:
* add delay options
* back off before retry when OverloadedException is raised

OverloadedException is raised by the underlying cassandra driver if the coordinator replies with an error code of OVERLOADED. this is an indication that the server is overloaded. but in general, this is an intermediate error, so there are chances that the same query could be served later on.

we could do the retry in the driver, but since the cassandra driver is implemented using netty, and the retry handling code is sitting in the netty's io thread. so we cannot just use `Thread.sleep()` in the io thread. it's no-go, just like performing a blocking call in Seastar's reactor thread. also, the retry machinery in the 3.x branch only offers 3 options:

- RETRY: either retry the request on current host, or switch to another
- RETHROW: throw the exception
- IGNORE: return an empty result

the RETRY one does not allow the retry policy to set `RetryDecision` with a configured delay. without a more invasive change, we cannot enable the driver to retry with a backoff when handling OverloadedException.

so, in this change, we add a couple options to allow cassandra-stress to back off when handling OverloadedException. this should increase the scylla's chance of completing a stress test even if the server returns OVERLOAD couple times when serving the queries from client before the configured retries are used up.

a similar change was proposed on cassandra jira 9 years ago, see https://issues.apache.org/jira/browse/CASSANDRA-7821. but it never made its way into the repo.

Refs https://github.com/scylladb/scylladb/issues/15192
Signed-off-by: Kefu Chai <kefu.chai@scylladb.com>

### DIFF
--- a/tools/stress/src/org/apache/cassandra/stress/Operation.java
+++ b/tools/stress/src/org/apache/cassandra/stress/Operation.java
@@ -21,6 +21,8 @@ package org.apache.cassandra.stress;
 import java.io.IOException;
 import java.util.NoSuchElementException;
 
+import com.datastax.driver.core.exceptions.OverloadedException;
+
 import org.apache.cassandra.stress.report.Timer;
 import org.apache.cassandra.stress.settings.SettingsLog;
 import org.apache.cassandra.stress.settings.StressSettings;
@@ -89,6 +91,19 @@ public abstract class Operation
             catch (NoSuchElementException e) {
                 // Pass thru iterator exhaustion exception
                 throw e;
+            }
+            catch (OverloadedException e) {
+                // The number of in-flight hints currently being written on the
+                // coordinator exceeds the limit, so we need to back off
+                try
+                {
+                    if (settings.log.level.compareTo(SettingsLog.Level.MINIMAL) > 0) {
+                        System.err.println(String.format("Server is overloaded, retry %d/%d times",
+                                                         tries, settings.errors.tries));
+                    }
+                    Thread.sleep(settings.errors.nextDelay(tries).toMillis());
+                }
+                catch (InterruptedException ie) { }
             }
             catch (Exception e)
             {
--- a/tools/stress/src/org/apache/cassandra/stress/settings/SettingsErrors.java
+++ b/tools/stress/src/org/apache/cassandra/stress/settings/SettingsErrors.java
@@ -23,8 +23,11 @@ package org.apache.cassandra.stress.settings;
 
 import java.io.Serializable;
 import java.util.Arrays;
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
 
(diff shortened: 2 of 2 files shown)


tchaikov:
@avelanarius hi Piotr, would you mind taking a look?

tchaikov:
@avelanarius hi Piotr, ping?

bhalevy:
> @avelanarius hi Piotr, ping?

@avelanarius ping.  please review

bhalevy:
> > @avelanarius hi Piotr, ping?
> 
> @avelanarius ping. please review

@avelanarius ping^. please review

avikivity:
Does a retry after OVERLOADED pick a different coordinator?

tchaikov:
> Does a retry after OVERLOADED pick a different coordinator?

@avikivity  i think this is a behavior of the java driver. and it depends on the retry policy, if the java driver is using the default retry policy, when a `OverloadedException` is raised, it retries with the next host. see https://github.com/scylladb/java-driver/blob/085dd34ef46b1de5111629bdccfe4056c4355676/driver-core/src/main/java/com/datastax/driver/core/policies/DefaultRetryPolicy.java#L141-L149 . but in the test, we were using a single node.

roydahan (approved):
LGTM.

roydahan:
@avelanarius please review or assign.

bhalevy:
ping.  @Bouncheck please review.
This blocks a fix for dtest/ci stability

Bouncheck:
ACK, I'll look at this

Bouncheck (approved):
LGTM.
Optional change I'd consider:
OverloadedException catch escapes logging - it may be useful to add some logging either where it's being caught or in nextDelay (or do `e instanceof OverloadedException` in last clause instead) so that we'll be able to see in logs that we're hitting a lot of overloads in case they happen.

tchaikov:
@Bouncheck i added a logging message when handling `OverloadedException`. does it look better now?

Bouncheck:
Yes, looks better now.

tchaikov:
@scylladb/scylla-maint hello maintainers, could you help queue this change?

denesb:
We have a merge freeze in effect.

tchaikov:
ahh, thought scylla-tools-java was not part of the lockdown..

tchaikov:
@mykaul yes, it was, and it still is. this change is indeed a fix intended to address the CI stability issue at https://github.com/scylladb/scylladb/issues/15192 9 months ago during a code freeze.

denesb:
Ok, CI related improvements are allowed indeed, you should have told me that this is such, I didn't guess based on the subject and description.

tchaikov:
> Ok, CI related improvements are allowed indeed, you should have told me that this is such, I didn't guess based on the subject and description.

will do next time. actually, i forgot why i created this fix when asking for the help from maintainers.  just thrilled at seeing Wojciech approval after waiting for months..

fruch:
@CodeLieutenant 

FYI

We probably want to pull this one into the new repo for c-s

PR #354, opened 2023-11-15
dist/redhat: s/jre-8-headless/jre-11-headless/

tchaikov:
in order to tighten the runtime dependencies on different platforms and different components, also, it is better to use a newer and better supported jre, let's use jre-11 in the rpm packaging as well. we build rpm and deb from the same tarball. and we build for both jre-8 and jre-11, so it should be safe to use jre-11 as the dependency on rpm based distros.


elcallio (approved):

syuu1228 (approved):
LGTM

denesb (approved):

denesb:
Submodule update: 935065fd8ddf359b8b22b92e025c16e24c362a24

PR #355, opened 2023-11-20
NodeProbe: allow addressing table name with colon in it

nyh:
As the ObjectName documentation explains, when an addressing an MBean in JMX, when a value contains a comma, equals, colon or quote - it must be quoted, otherwise the ObjectName cannot be constructed.

CQL or Alternator table names cannot contain any of these characters, but in Alternator the names of secondary indexes *do* contain a colon, so before this patch nodetool cannot address them - it tries to construct an ObjectName with that name, and fails.

Note that "nodetool tablestats" is also broken by this bug, because even if the user doesn't try to specify one specific table, nodetool iterates on all tables and fails when it reaches the view with the colon in its name.

So this patch adds the missing quoting if the table name contains a colon.  We don't quote table names that don't have a colon to preserve backward compatibility with old implmentation of JMX that don't know how to unquote these names. A separate patch to scylla-jmx added support for unquoting - see https://github.com/scylladb/scylla-jmx/pull/227.

After this patch "nodetool tablestats" can work even if there is an Alternator GSI or LSI.

### DIFF
--- a/src/java/org/apache/cassandra/tools/NodeProbe.java
+++ b/src/java/org/apache/cassandra/tools/NodeProbe.java
@@ -1460,6 +1460,13 @@ public class NodeProbe implements AutoCloseable
                 // Do we need some other way to address indexes?
                 //String type = cf.contains(".") ? "IndexTable" : "Table";
                 String type = "Table";
+                if (cf.contains(":")) {
+                    // If cf name contains a colon (as happens in Alternator
+                    // GSI) it needs to be quoted otherwise it can't be stored
+                    // in an ObjectName. Only Scylla's fork of the JMX knows
+                    // how to unquote it later, so only quote if necessary.
+                    cf = ObjectName.quote(cf);
+                }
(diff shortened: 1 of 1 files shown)


nyh:
See also:
1. https://github.com/scylladb/scylla-jmx/pull/227 - patch for scylla-jmx to support the quoted table names that this patch can send for a table with a colon in its name.

denesb (approved):

denesb:
@nyh should I wait for https://github.com/scylladb/scylla-jmx/pull/227 to promote in scylla.git, or can I merge right away?

mykaul:
> @nyh should I wait for [scylladb/scylla-jmx#227](https://github.com/scylladb/scylla-jmx/pull/227) to promote in scylla.git, or can I merge right away?

It's merged, I think we can merge this one as well.

denesb:
> > @nyh should I wait for [scylladb/scylla-jmx#227](https://github.com/scylladb/scylla-jmx/pull/227) to promote in scylla.git, or can I merge right away?
> 
> It's merged, I think we can merge this one as well.

It is only queued, and it will probably need dequeuing because it is apparently causing problems.

denesb:
Submodule update PR: https://github.com/scylladb/scylladb/pull/16146

Issue #356, opened 2023-11-28
[cassandra-stress] creation code of the schema isn't safe 


fruch:
## Issue description

- [ ] This issue is a regression.
- [x] It is unknown if this issue is a regression.

when multiple machines are using the same command we can end up in situation some of the command would fail like this:
```
java.lang.RuntimeException: Encountered exception creating schema
	at org.apache.cassandra.stress.settings.SettingsSchema.createKeySpacesNative(SettingsSchema.java:100)
	at org.apache.cassandra.stress.settings.SettingsSchema.createKeySpaces(SettingsSchema.java:69)
	at org.apache.cassandra.stress.settings.StressSettings.maybeCreateKeyspaces(StressSettings.java:228)
	at org.apache.cassandra.stress.StressAction.run(StressAction.java:58)
	at org.apache.cassandra.stress.Stress.run(Stress.java:143)
	at org.apache.cassandra.stress.Stress.main(Stress.java:62)
Caused by: com.datastax.driver.core.exceptions.InvalidQueryException: Keyspace 'keyspace1' does not exist
	at com.datastax.driver.core.exceptions.InvalidQueryException.copy(InvalidQueryException.java:49)
	at com.datastax.driver.core.DriverThrowables.propagateCause(DriverThrowables.java:35)
	at com.datastax.driver.core.DefaultResultSetFuture.getUninterruptibly(DefaultResultSetFuture.java:293)
	at com.datastax.driver.core.AbstractSession.execute(AbstractSession.java:58)
	at org.apache.cassandra.stress.util.JavaDriverClient.execute(JavaDriverClient.java:190)
	at org.apache.cassandra.stress.settings.SettingsSchema.createKeySpacesNative(SettingsSchema.java:86)
	... 5 more
Caused by: com.datastax.driver.core.exceptions.InvalidQueryException: Keyspace 'keyspace1' does not exist
	at com.datastax.driver.core.Responses$Error.asException(Responses.java:181)
	at com.datastax.driver.core.DefaultResultSetFuture.onSet(DefaultResultSetFuture.java:215)
	at com.datastax.driver.core.RequestHandler.setFinalResult(RequestHandler.java:229)
	at com.datastax.driver.core.RequestHandler.access$2600(RequestHandler.java:63)
	at com.datastax.driver.core.RequestHandler$SpeculativeExecution.setFinalResult(RequestHandler.java:1011)
[...]

PR #358, opened 2023-11-30
cassandra-stress: switch create table to use keyspace.table format

roydahan:
Stop using "Use keyspace" since it can fail when using multiple stress commands in parallel due to a race condition.

fixes: #356

### DIFF
--- a/tools/stress/src/org/apache/cassandra/stress/settings/SettingsSchema.java
+++ b/tools/stress/src/org/apache/cassandra/stress/settings/SettingsSchema.java
@@ -87,9 +87,7 @@ public class SettingsSchema implements Serializable
         {
             //Keyspace
             client.execute(createKeyspaceStatementCQL3(), org.apache.cassandra.db.ConsistencyLevel.LOCAL_ONE);
-
-            client.execute("USE \""+keyspace+"\"", org.apache.cassandra.db.ConsistencyLevel.LOCAL_ONE);
-
+            
             //Add standard1
             client.execute(createStandard1StatementCQL3(settings), org.apache.cassandra.db.ConsistencyLevel.LOCAL_ONE);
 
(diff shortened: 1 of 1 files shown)


tools/stress/src/org/apache/cassandra/stress/settings/SettingsSchema.java
```diff
@@ -152,7 +150,7 @@ String createStandard1StatementCQL3(StressSettings settings)
         StringBuilder b = new StringBuilder();
 
         b.append("CREATE TABLE IF NOT EXISTS ")
-         .append("standard1 (key blob PRIMARY KEY ");
+         .append("\""+keyspace+"\"".standard1 (key blob PRIMARY KEY ");
```
fruch, line 153: `createCounter1StatementCQL3` might be missing it as well, even that I didn't found any place in the code that calls it.

mykaul:
ping

roydahan:
I never had the chance to test it, but it's also very rare case, so not urgent.

avelanarius:
Does it really solve #356? I don't see why it'd solve that problem - if `USE keyspace` would fails, why wouldn't `CREATE keyspace.table` also fail?

roydahan:
The problem happens when you run multiple clients at the same time.
It's a race condition that would be hard to prove that is fixed.
However, In any case I don't see any benefit of using "USE keyspace" instead of creating with full path.

I still need to test that it doesn't break anything, just didn't find a time for it yet.

mykaul:
What's holding this?

avelanarius:
> What's holding this?

This PR lacks any explanation/proof why this really fixes the problem.

And personally I believe that this doesn't fix anything. If it really fixed the problem, that would mean that there is a serious bug in Java Driver around USE statements that should be fixed and not worked around.

fruch:
> > What's holding this?
> 
> This PR lacks any explanation/proof why this really fixes the problem.
> 
> And personally I believe that this doesn't fix anything. If it really fixed the problem, that would mean that there is a serious bug in Java Driver around USE statements that should be fixed and not worked around.

see https://github.com/scylladb/scylladb/pull/16969, it's a really issue in scylla's end, that create KS doesn't always return to the driver that the schema has changed.

anyhow this change is removing one command, that lower the changes of this issue to be happening

roydahan:
The real fix should be this one: https://github.com/scylladb/scylladb/pull/16969.
Closing this in favor of handling the real issue.

PR #361, opened 2023-12-08
Add nodetool interposer script

denesb:
This script is intended as a wrapper for nodetool, first trying to route commands to scylla-nodetool, and if the command is unimplemented, it is routed to the java nodetool. Whether a command is implemented or not is determined by checking the exit-code of scylla-nodetool. An unimplemented command uses a unique exit-code of 100. When this is seen, the command is routed to Java nodetool.
`install.sh` is adjusted to install `nodetool` as `nodetool.java` and `nodetool-wrapper` as `nodetool`. If scylla-nodetool is misbehaving, users can manually fall-back to the legacy nodetool, by invoking `nodetool.java`.

Tests: build both the unified tarball as well as the tools-java one and verified that the script works as intended at the installed location. This will also be checked by Scylla CI.

Refs: https://github.com/scylladb/scylladb/issues/15588

### DIFF
--- a/bin/nodetool-wrapper
+++ b/bin/nodetool-wrapper
@@ -1,0 +1,35 @@
+#!/bin/bash
+#
+# Copyright (C) 2023-present ScyllaDB
+#
+
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+#
+
+SCRIPT_PATH=$(dirname $(realpath "$0"))
+
+# nodetool is installed to $prefix/share/cassandra/bin, while scylla to $prefix/bin
+SCYLLA_PATH="$SCRIPT_PATH"/../../../bin/scylla
+NODETOOL_PATH=$(realpath "$SCRIPT_PATH"/nodetool-java)
+
+# Fall-back to java nodetool, if the tool-java package is installed, but the
+# scylla-server one isn't.
+if [[ ! -f "$SCYLLA_PATH" ]]
+then
+    exec "$NODETOOL_PATH" "$@"
+fi
+
+SCYLLA_PATH=$(realpath "$SCYLLA_PATH")
+
+# Requesting the help for a command will fail with a distinct error code (100)
+# if the command doesn't exist. Thus we can use the exit code to check whether
+# scylla-nodetool implements a given command or not, and therefore, dispatch
+# the request to the appropriate nodetool variant.
+"$SCYLLA_PATH" nodetool "$1" --help &>/dev/null
+if [[ $? -eq 100 ]]
+then
+    exec "$NODETOOL_PATH" "$@"
+else
+    exec "$SCYLLA_PATH" nodetool "$@"
+fi
--- a/install.sh
+++ b/install.sh
@@ -117,14 +117,19 @@ if $nonroot; then
     sed -i -e "s#/opt/scylladb/#$rprefix/#g" "$rprefix"/share/cassandra/bin/cassandra.in.sh
 fi
 
-# scylla-tools
-install -d -m755 "$retc"/bash_completion.d
-install -m644 dist/common/nodetool-completion "$retc"/bash_completion.d
-for i in bin/{nodetool,sstableloader,scylla-sstableloader} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
-    bn=$(basename $i)
-    install -m755 $i "$rprefix"/share/cassandra/bin
+install_tool_bin () {
+    bin_src=$1
+    bn=$(basename "$bin_src")
+    install -m755 "$bin_src" "$rprefix"/share/cassandra/bin
     if ! $nonroot; then
         echo "$thunk" > "$rusr"/bin/$bn
         chmod 755 "$rusr"/bin/$bn
     fi
+}
+
+# scylla-tools
+install -d -m755 "$retc"/bash_completion.d
+install -m644 dist/common/nodetool-completion "$retc"/bash_completion.d
+for i in bin/{nodetool,sstableloader,scylla-sstableloader} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
+    install_tool_bin "$i"
 done
--- a/install.sh
+++ b/install.sh
@@ -134,6 +134,9 @@ install_tool_bin () {
 # scylla-tools
 install -d -m755 "$retc"/bash_completion.d
 install -m644 dist/common/nodetool-completion "$retc"/bash_completion.d
-for i in bin/{nodetool,sstableloader,scylla-sstableloader} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
+for i in bin/{sstableloader,scylla-sstableloader} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
     install_tool_bin "$i"
 done
+
+install_tool_bin bin/nodetool-wrapper nodetool
+install_tool_bin bin/nodetool nodetool-java
--- a/install.sh
+++ b/install.sh
@@ -119,11 +119,15 @@ fi
 
 install_tool_bin () {
     bin_src=$1
-    bn=$(basename "$bin_src")
-    install -m755 "$bin_src" "$rprefix"/share/cassandra/bin
+    if [ $# -eq 1 ]; then
+        bin_dest=$(basename "$bin_src")
+    else
+        bin_dest=$2
+    fi
+    install -m755 "$bin_src" "$rprefix"/share/cassandra/bin/"$bin_dest"
     if ! $nonroot; then
-        echo "$thunk" > "$rusr"/bin/$bn
-        chmod 755 "$rusr"/bin/$bn
+        echo "$thunk" > "$rusr"/bin/"$bin_dest"
+        chmod 755 "$rusr"/bin/"$bin_dest"
     fi
 }
 
--- a/dist/debian/debian/scylla-tools.install
+++ b/dist/debian/debian/scylla-tools.install
@@ -1,5 +1,6 @@
 etc/bash_completion.d/nodetool-completion
 opt/scylladb/share/cassandra/bin/nodetool
+opt/scylladb/share/cassandra/bin/nodetool-java
 opt/scylladb/share/cassandra/bin/sstableloader
 opt/scylladb/share/cassandra/bin/cassandra-stress
 opt/scylladb/share/cassandra/bin/cassandra-stressd
@@ -8,6 +9,7 @@ opt/scylladb/share/cassandra/bin/sstablelevelreset
 opt/scylladb/share/cassandra/bin/sstablemetadata
 opt/scylladb/share/cassandra/bin/sstablerepairedset
 usr/bin/nodetool
+usr/bin/nodetool-java
 usr/bin/sstableloader
 usr/bin/cassandra-stress
 usr/bin/cassandra-stressd
--- a/dist/redhat/scylla-tools.spec
+++ b/dist/redhat/scylla-tools.spec
@@ -42,6 +42,7 @@ rm -rf $RPM_BUILD_ROOT
 %files
 %{_sysconfdir}/bash_completion.d/nodetool-completion
 /opt/scylladb/share/cassandra/bin/nodetool
+/opt/scylladb/share/cassandra/bin/nodetool-java
 /opt/scylladb/share/cassandra/bin/sstableloader
 /opt/scylladb/share/cassandra/bin/cassandra-stress
 /opt/scylladb/share/cassandra/bin/cassandra-stressd
@@ -50,6 +51,7 @@ rm -rf $RPM_BUILD_ROOT
 /opt/scylladb/share/cassandra/bin/sstablemetadata
 /opt/scylladb/share/cassandra/bin/sstablerepairedset
 %{_bindir}/nodetool
+%{_bindir}/nodetool-java
 %{_bindir}/sstableloader
 %{_bindir}/cassandra-stress
 %{_bindir}/cassandra-stressd


nyh:
I'm curious, why did you want this script in scylla-tools-java.git and not scylla.git?
The benefit of putting it in scylla.git is that you can later improve it (or maybe even remove it!) in the same patch with a Scylla patch, and won't need those ugly two-repository dances.
But I don't insist, if you like it this way.

nyh:
> I'm curious, why did you want this script in scylla-tools-java.git and not scylla.git? The benefit of putting it in scylla.git is that you can later improve it (or maybe even remove it!) in the same patch with a Scylla patch, and won't need those ugly two-repository dances. But I don't insist, if you like it this way.

Oh, I misunderstood that you wanted the new script to hide the old nodetool, and other projects (e.g., dtest) which used the old nodetool will start to use yours. Then I guess I understand why it needs to be in the nodetool repository. So sorry about the spam.

bin/nodetool-wrapper
```diff
@@ -0,0 +1,29 @@
...
+#
+
+SCRIPT_PATH="$(dirname $(realpath $0))"
+
+# nodetool is installed to $prefix/share/cassandra/bin, while scylla to $prefix/bin
+SCYLLA_PATH="$(realpath $SCRIPT_PATH/../../../bin/scylla)"
+NODETOOL_PATH="$(realpath $SCRIPT_PATH/nodetool.java)"
+
+# Capture the output of scylla-nodetool, as it might just print an error about
+# the command not existing.
```
nyh, line 17: An idea how to improve this (I think I mentioned this in the past, it's not important to do this now but I think it will be a good idea for you to add alter) -

I think you should add a command line option, e.g., "-q" (quiet) or "-s" (silent), which would cause "scylla nodetool" to be silent if the command is not supported and JUST return the error code 100. This will allow you to get rid of this ugly redirection trick, which isn't just ugly (sorry), it's also partially wrong because it:
1. Can buffer output longer than scylla-nodetool wanted to.
2. Will combine both stderr to stdout into one stdout stream.

nyh (approved):
Looks good. I have an idea how to improve the interposer to get rid of the "ugly" redirection trick, but it can be merged as-is and improved (if you want) later.

tchaikov:
for the posterity, this change was previously implemented in scylladb, see https://github.com/scylladb/scylladb/pull/16087

bin/nodetool-wrapper
```diff
@@ -0,0 +1,29 @@
...
+# Only print $OUT after checking the return-code for the specific unimplemented
+# command error code.
+OUT=$($SCYLLA_PATH nodetool "$@" 2>&1)
```
tchaikov, line 20: not sure if it is a valid concern, do we have use case where our test / script / workflow capture the stdout and stderr into different variables? if it's true, we might want to reconsider the comment at https://github.com/scylladb/scylladb/pull/16087#discussion_r1399664492 and above.

tchaikov (commented):
left a couple comments.

bin/nodetool-wrapper
nyh: Right - I knew I thought about this already :-)
I prefer my proposed solution ("-q") over your "--help" one, because in the optimal case (90% of the nodetool runs will just use Scylla) it will just run scylla once, not twice.

install.sh
```diff
@@ -117,14 +117,19 @@ if $nonroot; then
...
 fi
 
-# scylla-tools
-install -d -m755 "$retc"/bash_completion.d
-install -m644 dist/common/nodetool-completion "$retc"/bash_completion.d
-for i in bin/{nodetool,sstableloader,scylla-sstableloader} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
-    bn=$(basename $i)
-    install -m755 $i "$rprefix"/share/cassandra/bin
+install_tool_bin () {
+    bn_src=$1
```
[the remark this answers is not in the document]
nyh: I think "bn" was supposed to be shorthand for basename, not bin?
Anyway, it seems this "bn" name was used in the existing version, so he just kept it?

bin/nodetool-wrapper
nyh: Yes, I don't know if in practice nodetool prints different things to stdout vs stderr, or if sometimes it prints unbuffered output (e.g., consider a long-running command that prints output incrementally), but in principle - it may - these are valid things a Unix / Linux command might do. And unless we have a very good reason, we shouldn't break it. And I think we don't need to break it if we only add the "-q" option as I suggested above.

install.sh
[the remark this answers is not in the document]
tchaikov: sorry, i withdraw my comment on the `bin` part. indeed, `bn` should be the short for `basename`.

bin/nodetool-wrapper
denesb: If this was a longer-term solution, I would prefer the `-q/-s` option. But since I plan to wrap the transition up in a few months, I don't want to add an option, which doesn't make sense outside the scope of this script (a user always wants readable error messages, instead of silent failures).
We can live with the extra `--help` call for this transitional period.

denesb:
New in v2:
* install.sh: 
    - make style more consistent with rest of the file (`then` on same line as `if`).
    - `install_tool_bin()`: use more descriptive variable names
    - more quoting to make bash-linter happy
* nodetool-wrapper: use `scylla nodetool $COMMAND --help` to poll for command implementation, this way we don't need games with output capture.

bin/nodetool-wrapper
denesb: Got rid of this in v2.

install.sh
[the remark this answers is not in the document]
denesb: Used better names in v2.

tchaikov (commented):
lgtm.

bin/nodetool-wrapper
denesb: Yes, we will have this script for some time, unless we backport scylla-nodetool patches (which I doubt).
Note that implementing this `-q` would not be trivial. We lookup the command and check for `--help` very early in the startup phase. At this point, the regular arguments are not even parsed yet. So I'd have to shuffle the startup code around to add support to this special `-q` flag and the only thing it would change is that I wouldn't need a redirection to `/dev/null`, which is not a large gain.

As for the implementation state, currently, we stand at 31/48 commands, so 17 more commands to go. I already implemented all the low-hanging fruits, still I think I will likely finish by the 6.0 timeline (although I don't promise).

denesb: I implemented @tchaikov's suggestion in v2. It indeed makes the code much nicer.

install.sh
```diff
@@ -117,14 +117,27 @@ if $nonroot; then
...
+    if [ $# -eq 1 ]
+    then
```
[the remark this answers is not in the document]
denesb: Done in v2.

install.sh
```diff
@@ -117,14 +117,27 @@ if $nonroot; then
...
+        bn=$(basename $bn_src)
```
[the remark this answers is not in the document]
denesb: Done in v2.

bin/nodetool-wrapper
denesb: I just realized that with `-q` we would only need to run `scylla-nodetool` once, instead of twice. I don't think this makes a big difference, `scylla-nodetool` starts very fast, and the slow path (scylla-nodetool + java-nodetool) still remains, with both solutions.

nyh: I'm not sure what you mean. With "-q" you run it just once - it either does the right thing, or in the case of missing support (and only in that case) it fails **silently**. I think with @tchaikov's "--help" suggestion you need to run it twice - once just to check the validity, and again to run normally. This is why I liked this idea less. But I agree that if scylla-nodetool starts very fast (surprising, considering it is probably a huge executable which uses Seastar) it doesn't matter.

denesb:
CI PR: https://github.com/scylladb/scylladb/pull/16366

install.sh
```diff
@@ -117,14 +117,19 @@ if $nonroot; then
...
 
-# scylla-tools
-install -d -m755 "$retc"/bash_completion.d
-install -m644 dist/common/nodetool-completion "$retc"/bash_completion.d
-for i in bin/{nodetool,sstableloader,scylla-sstableloader} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
-    bn=$(basename $i)
-    install -m755 $i "$rprefix"/share/cassandra/bin
+install_tool_bin () {
+    bin_src=$1
+    bn=$(basename "$bin_src")
```
avikivity: Quotes around the right-hand side are not strictly necessary, but fear is the right approach when dealing with bash.

install.sh
```diff
@@ -117,14 +117,19 @@ if $nonroot; then
...
+    install -m755 "$bin_src" "$rprefix"/share/cassandra/bin
     if ! $nonroot; then
         echo "$thunk" > "$rusr"/bin/$bn
```
avikivity: Here quotes ($bn) are necessary.

bin/nodetool-wrapper
```diff
@@ -0,0 +1,26 @@
...
+
+# nodetool is installed to $prefix/share/cassandra/bin, while scylla to $prefix/bin
+SCYLLA_PATH=$(realpath "$SCRIPT_PATH"/../../../bin/scylla)
+NODETOOL_PATH=$(realpath "$SCRIPT_PATH"/nodetool.java)
+
+# Requesting the help for a command will fail with a distinct error code (100)
+# if the command doesn't exist. Thus we can use the exit code to check whether
+# scylla-nodetool implements a given command or not, and therefore, dispatch
+# the request to the appropriate nodetool variant.
+$SCYLLA_PATH nodetool "$1" --help &>/dev/null
```
avikivity, line 20: First, quotes around SCYLLA_PATH. Second, need to check that it in fact exists, scylla-tools-java may have been installed on a non-server node (happens often to get cassandra-stress, I don't believe anyone uses it for nodetool, but let's not break it).

avikivity: Also, "$1" may not be the command, there could be options to nodetool.

bin/nodetool-wrapper
```diff
@@ -0,0 +1,26 @@
...
+if [[ $? -eq 100 ]]
+then
+    exec $NODETOOL_PATH "$@"
+else
+    exec $SCYLLA_PATH nodetool "$@"
```
avikivity, line 25: quotes quotes

mykaul:
Instead of running --help you could have run it once on installation, save it somewhere and read from there. I think it's a micro-optimization that isn't worth it though.

install.sh
```diff
@@ -134,6 +134,9 @@ install_tool_bin () {
 # scylla-tools
 install -d -m755 "$retc"/bash_completion.d
 install -m644 dist/common/nodetool-completion "$retc"/bash_completion.d
-for i in bin/{nodetool,sstableloader,scylla-sstableloader} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
+for i in bin/{sstableloader,scylla-sstableloader} tools/bin/{cassandra-stress,cassandra-stressd,sstabledump,sstablelevelreset,sstablemetadata,sstablerepairedset}; do
     install_tool_bin "$i"
 done
+
+install_tool_bin bin/nodetool-wrapper nodetool
+install_tool_bin bin/nodetool nodetool.java
```
avikivity, line 142: I know I recommended nodetool.java, but it was a bad choice, that's used for Java source files and it would just confuse people even more than they are now. How about nodetool-java.

avikivity: Do the deb/rpm scripts work? They may, if they use wildcards to capture the contents of bin/.

avikivity:
> Instead of running --help you could have run it once on installation, save it somewhere and read from there. I think it's a micro-optimization that isn't worth it though.

It's a micro-optimization but adds macro-complexity. Where do you store it? How do you know when to refresh it? Do you have write permissions? And in any case the whole thing disappears once we figure out what N is in 5/N.

This is just a way to get advance testing before the whole thing is complete.

avikivity:
Please test with `./install.sh --prefix ~/tmp123` from the tarball, just to be sure. I think it will work with no changes (well it might not work if "tmp123" has spaces injected).

denesb:
> Please test with `./install.sh --prefix ~/tmp123` from the tarball, just to be sure. I think it will work with no changes (well it might not work if "tmp123" has spaces injected).

I already did test it like this. I guess I should have mentioned this in the patch log.

install.sh
denesb: I added in the next patch, which replaces `$bn`.

bin/nodetool-wrapper
denesb: Good point.

install.sh
denesb: > I know I recommended nodetool.java, but it was a bad choice, that's used for Java source files and it would just confuse people even more than they are now. How about nodetool-java.

Will rename.

> Do the deb/rpm scripts work? They may, if they use wildcards to capture the contents of bin/.

Will check.

denesb:
```
13:21:17  error: Installed (but unpackaged) file(s) found:
13:21:17     /opt/scylladb/share/cassandra/bin/nodetool.java
13:21:17     /usr/bin/nodetool.java
13:21:17  
13:21:17  RPM build errors:
13:21:17      Installed (but unpackaged) file(s) found:
13:21:17     /opt/scylladb/share/cassandra/bin/nodetool.java
13:21:17     /usr/bin/nodetool.java
```
I guess we have our answer. I will need to fix this too.

denesb:
New in v3:
* nodetool-wrapper:
    - more quoting
    - check if $SCYLLA_PATH exists, fall-back to java nodetool if not
*  s/nodetool.java/nodetool-java/
* dist: add nodetool-java to rpm and deb specs

install.sh
denesb: nodetool.java renamed to nodetool-java in v3
Updated dist/ in v3.

bin/nodetool-wrapper
denesb: Done in v3.

bin/nodetool-wrapper
denesb: Done in v3.

denesb:
New in v3.1:
* nodetool-wrapper: don't use `realpath` on the `SCYLLA_PATH` until we don't know it exists. It results in an ugly error message printed.

denesb:
We have green light from CI: https://github.com/scylladb/scylladb/pull/16366#issuecomment-1851894236.

nyh:
I merged this patch to scylla-tools-java, but we also need to do the submodule update in Scylla (that's https://github.com/scylladb/scylladb/pull/16366, I guess?).

avikivity:
@nyh don't half-merge things, if you merge into the submodule then also update the submodule reference

nyh:
@avikivity as I said, the second half of the merge was https://github.com/scylladb/scylladb/pull/16366 - a PR that @denesb already opened. If he hadn't, I would have to do it myself (to test the submodule update before merging it), but he already did.

denesb:
> @avikivity as I said, the second half of the merge was [scylladb/scylladb#16366](https://github.com/scylladb/scylladb/pull/16366) - a PR that @denesb already opened. If he hadn't, I would have to do it myself (to test the submodule update before merging it), but he already did.

Said PR was opened for CI only. I wanted to get CI feedback while review was ongoing, as I was expecting some problems.
My idea was that once we have CI pass, there is no need for the submodule update for this PR to go through another scylla.git PR.

Sorry for the confusion.

nyh:
> @nyh it's not generally possible to open pull requests for submodule updates, because submodule history is linear and does not tolerate merges. The maintainer should perform the merge and submodule update close together to avoid another merge from updating the submodule under its feet.

The latter is true, but PRs are still the best way to go (even if you have to merge them quickly), especially in submodules like Seastar where a submodule refresh can bring in a lot of unrelated changes, and they should all be tested to not break next.

avikivity:
So far the procedure is for the maintainer to generate the submodule update.

PR #362, opened 2023-12-11
throw more informative error when fail to parse sstable generation

### LINKED ISSUE #360: patch java tools print a message when they see a UUID generation
since scylla 5.4, the sstables are named with a uuid-based identifier instead of integer-based one. so the java tools inherited from cassandra 3.x branch do not support it, and throw exception, like
```
root@df13a4d04afa:/var/lib/scylla/data/foo/bar-eb90daf0952d11ee87e220c83109c794# sstablemetadata 'me-3gbp_1glu_4e6g020ns4px173el0-big-Data.db'
Exception in thread "main" java.lang.NumberFormatException: For input string: "3gbp_1glu_4e6g020ns4px173el0"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:652)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:280)
	at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:228)
	at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:217)
	at org.apache.cassandra.tools.SSTableMetadataViewer.main(SSTableMetadataViewer.java:150)
root@df13a4d04afa:/var/lib/scylla/data/foo/bar-eb90daf0952d11ee87e220c83109c794# sstabledump 'me-3gbp_1glu_4e6g020ns4px173el0-big-Data.db'
Exception in thread "main" java.lang.NumberFormatException: For input string: "3gbp_1glu_4e6g020ns4px173el0"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:652)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:280)
	at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:228)
	at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:217)
	at org.apache.cassandra.tools.SSTableExport.run(SSTableExport.java:188)
	at com.scylladb.tools.SSTableExport.main(SSTableExport.java:44)
```

we should patch java tools print a message when they see a UUID generation, telling the user to use the native tools for sstables with UUID generation.

tchaikov:
since scylla 5.4, the sstables are named with a uuid-based identifier instead of integer-based one. so the java tools inherited from cassandra 3.x branch do not support it, and throw an
`UnsupportedOperationException` when parsing a sstable name with the integer-based generation. like:

```console
$ sstablemetadata me-3gbp_1glu_4e6g020ns4px173el0-big-Data.db
Exception in thread "main" java.lang.NumberFormatException: For input string: "3gbp_1glu_4e6g020ns4px173el0"
        at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.base/java.lang.Integer.parseInt(Integer.java:652)
        at java.base/java.lang.Integer.parseInt(Integer.java:770)
        at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:280)
        at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:228)
        at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:217)
        at org.apache.cassandra.tools.SSTableMetadataViewer.main(SSTableMetadataViewer.java:150)
```

the user experience is poor when seeing this exception. so let's throw a better exception, so the new output looks like:

```console
$ sstablemetadata me-3gbr_0y9h_03n682utcozh4xc6c7-big-Data.db
Exception in thread "main" java.lang.UnsupportedOperationException: SSTable 'me-3gbr_0y9h_03n682utcozh4xc6c7-big-Data.db' is using the UUID-based identifier, and hence should be openned using 'scylla sstable' commands
        at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:295)
        at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:228)
        at org.apache.cassandra.io.sstable.Descriptor.fromFilename(Descriptor.java:217)
        at org.apache.cassandra.tools.SSTableMetadataViewer.main(SSTableMetadataViewer.java:150)
```

Fixes #360
Signed-off-by: Kefu Chai <kefu.chai@scylladb.com>


src/java/org/apache/cassandra/io/sstable/Descriptor.java
```diff
@@ -277,7 +277,22 @@ public static Pair<Descriptor, String> fromFilename(File directory, String name,
...
+        {
+            // since Scylla 5.4, the SStable started using UUID v1 identifier
+            // instead of sequence of integers. but C* 3.x does not support the
+            // former. let's direct user to use the native tools of
+            // `scylla sstable` which is builtin in the scylla executable,
+            generation = Integer.parseInt(nexttok);
+        }
+        catch (NumberFormatException e)
+        {
+            throw new UnsupportedOperationException("SSTable '" + name
```
denesb, line 291: Can we throw some unique exception type here, that we catch in main and print only the message without a stacktrace. I think even though the message now is much better, the stracktrace is distracting, it would be preferable to omit it. If this would lead to a lot of churn, then we can keep this change instead.

tchaikov: sure. then we need to update every single tool.

tchaikov:
v2:

- catch and print out the exception in `main()`.

tchaikov:
@denesb hi Botond, could you take another look? the latest revision should have addressed your comment.

src/java/org/apache/cassandra/io/sstable/Descriptor.java
```diff
@@ -277,7 +277,22 @@ public static Pair<Descriptor, String> fromFilename(File directory, String name,
...
+                                                    + "' is using the UUID-based identifier, "
+                                                    + "and hence should be openned using "
+                                                    + "'scylla sstable' commands.");
```
fee-mendes, line 294: Perhaps just `use 'scylla sstable' operations instead.` is simpler and straight to the point.

tchaikov: thank you. i reword the error message a little bit, so it looks like:

> SSTable 'x.db' uses UUID-based identifier. Please use 'scylla sstable' commands instead.

1. `scylla sstable` represents a set of commands
2. add "Please" .

tchaikov:
v2:

- reword the error message. and update the commit message to match with the error message.

fee-mendes (approved):

PR #367, opened 2023-12-14
Use ReplicaOrdering.NEUTRAL in TokenAwarePolicy to respect RackAwareness

sylwiaszunejko:
If we specify `rack`, we want to have replicas sorted based on which one is local. This happens on the java-driver side if we specify ordering in `TokenAwarePolicy`. Specifically, it has to use `ReplicaOrdering.NEUTRAL`. 

This option was not used before, and that led to broken behavior. Now, Rack Awareness is properly respected.

Fixes: https://github.com/scylladb/java-driver/issues/255 

To test I have used case with 6 nodes, and c-s is using `RackAwareRoundRobinPolicy` that targets only rack `rack1`
and all stress commands are showing the following logs:
```
===== Using optimized driver!!! =====
WARN  13:57:02,324 Some contact points don't match local rack. Local rack = rack1. Non-conforming contact points: /127.0.0.2:9042 (dc=datacenter1, rack=rack2),/127.0.0.3:9042 (dc=datacenter1, rack=rack3),/127.0.0.5:9042 (dc=datacenter1, rack=rack2),/127.0.0.6:9042 (dc=datacenter1, rack=rack3)
Connected to cluster: , max pending requests per connection null, max connections per host 8
Datatacenter: datacenter1; Host: /127.0.0.1; Rack: rack1
Datatacenter: datacenter1; Host: /127.0.0.2; Rack: rack2
Datatacenter: datacenter1; Host: /127.0.0.3; Rack: rack3
Datatacenter: datacenter1; Host: /127.0.0.4; Rack: rack1
Datatacenter: datacenter1; Host: /127.0.0.5; Rack: rack2
Datatacenter: datacenter1; Host: /127.0.0.6; Rack: rack3
```
Before the change for `rf=3` it looked like that (first write operation, after that some reads):
rack_before
After the change for the same parameters we see that majority of the requests goes to `127.0.0.1` and `127.0.0.4` (orange and green lines) as it should.
rack_after

### DIFF
--- a/tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
+++ b/tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
@@ -27,6 +27,7 @@ import com.datastax.driver.core.*;
 import com.datastax.driver.core.policies.RackAwareRoundRobinPolicy;
 import com.datastax.driver.core.policies.LoadBalancingPolicy;
 import com.datastax.driver.core.policies.TokenAwarePolicy;
+import com.datastax.driver.core.policies.TokenAwarePolicy.ReplicaOrdering;
 import com.datastax.driver.core.policies.WhiteListPolicy;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 import io.netty.util.internal.logging.Slf4JLoggerFactory;
@@ -107,7 +108,7 @@ public class JavaDriverClient
         if (settings.node.isWhiteList)
             ret = new WhiteListPolicy(ret == null ? policyBuilder.build() : ret, settings.node.resolveAll(settings.port.nativePort));
 
-        return new TokenAwarePolicy(ret == null ? policyBuilder.build() : ret);
+        return new TokenAwarePolicy(ret == null ? policyBuilder.build() : ret, ReplicaOrdering.NEUTRAL);
     }
 
     public PreparedStatement prepare(String query)


avelanarius (approved):

sylwiaszunejko:
@avelanarius I added some graphs to show how the change affects them.

roydahan:
@nyh / @denesb can you please review and merge if approved?

BTW, what's the process to ping scylla maintainers and how would one get a PR merged without pinging maintainers?

tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
```diff
@@ -107,7 +108,7 @@ private LoadBalancingPolicy loadBalancingPolicy(StressSettings settings)
         if (settings.node.isWhiteList)
             ret = new WhiteListPolicy(ret == null ? policyBuilder.build() : ret, settings.node.resolveAll(settings.port.nativePort));
 
-        return new TokenAwarePolicy(ret == null ? policyBuilder.build() : ret);
+        return new TokenAwarePolicy(ret == null ? policyBuilder.build() : ret, ReplicaOrdering.NEUTRAL);
```
fruch, line 111: What is the default for that parameter ? why can the driver make this as the default ?

fruch: Also is the driver docs:
https://github.com/scylladb/java-driver/blob/085dd34ef46b1de5111629bdccfe4056c4355676/manual/load_balancing/README.md

Doesn't mention this at all

Just that shuffle is on by default

I think it should be highlighted in the RackAware part

fruch (approved):
Lgtm

Some clarifications on defaults and docs

tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
sylwiaszunejko: here you can see the default: https://github.com/scylladb/java-driver/blob/085dd34ef46b1de5111629bdccfe4056c4355676/driver-core/src/main/java/com/datastax/driver/core/policies/TokenAwarePolicy.java#L131

here is the description that shows the need to use ReplicaOrdering.NEUTRAL https://github.com/scylladb/java-driver/blob/085dd34ef46b1de5111629bdccfe4056c4355676/driver-core/src/main/java/com/datastax/driver/core/policies/TokenAwarePolicy.java#L90-L98

I will create an issue to add information about that to the docs.

fruch: I can say know drivers code (i.e. mostly the python-driver), I would have guessed that from current code or from any example using RackAware, 

so i'm guess any external user would have hard time figuring it out

mykaul:
> @nyh / @denesb can you please review and merge if approved?
> 
> BTW, what's the process to ping scylla maintainers and how would one get a PR merged without pinging maintainers?

Give them a call.
I think pinging the maintainers is a standard practice. (but I do see some unresolved comments here, that might confuse them)

tools/stress/src/org/apache/cassandra/stress/util/JavaDriverClient.java
sylwiaszunejko: Could we mark this comment as resolved or firstly you want this doc issue to be completed? @fruch

fruch: we could (but I don't have permission to resolve issue here)

sylwiaszunejko:
@mykaul all comments are resolved now

mykaul:
@scylladb/scylla-maint  (or is it @scylladb/scylla-jmx-maint  ?) - please consider merging.

denesb:
submodule update: scylladb/scylladb@f6f1ee70ed8df7aba3cfcb941b537fc85ec4da03

denesb:
> submodule update: [scylladb/scylladb@f6f1ee7](https://github.com/scylladb/scylladb/commit/f6f1ee70ed8df7aba3cfcb941b537fc85ec4da03)

I dequeued this, I forgot we cannot update tools-java until https://github.com/scylladb/java-driver/pull/267 is fixed.

roydahan:
https://github.com/scylladb/java-driver/pull/267 passed the CI and now merged.

mykaul:
> [scylladb/java-driver#267](https://github.com/scylladb/java-driver/pull/267) passed the CI and now merged.

Now we need to update the Java driver version in scylla-tools-java.

avelanarius:
Java Driver 3.11.5.1 is now released so the version number can be bumped in scylla-tools-java.

Issue #207, opened 2020-09-21
scylla-sstableloader gives Cannot generate <> comparison for primary key colum ColumnDefinition{name=h, type=org.apache.cassandra.db.marshal.Int32Type, kind=CLUSTERING, position=0}


pidiaquez:
Component: scylla-sstableloader

version:  scylla-enterprise-tools-core-2018.1.14-0.20190807.0cabdff9a4.el7.noarch

used lo load a table with 500G , jus 1 table ( DDL below )

data is coming from a snapshot taken from the same node.

just 1 table being loaded.  KS=phoenix_v3  table  [tenant_tag_data_points]

`CREATE TABLE phoenix_v3.tenant_tag_data_points (
tnt text,
thsh text,
l int,
ahsh text,
v int,
h int,
a text static,
d boolean,
g blob,
q tinyint,
t text static,
y tinyint,
PRIMARY KEY ((tnt, thsh, l, ahsh, v), h)
) WITH CLUSTERING ORDER BY (h ASC)`

after scylla-sstableloader works loading records for 2h is finish with exception  

`java.lang.IllegalStateException: Cannot generate <> comparison for primary key colum ColumnDefinition{name=h, type=org.apache.cassandra.db.marshal.Int32Type, kind=CLUSTERING, position=0}
at com.scylladb.tools.SSTableToCQL$RowBuilder.setWhere(SSTableToCQL.java:335)
at com.scylladb.tools.SSTableToCQL$RowBuilder.deleteCqlRow(SSTableToCQL.java:382)
at com.scylladb.tools.SSTableToCQL$RowBuilder.process(SSTableToCQL.java:728)
at com.scylladb.tools.SSTableToCQL$RowBuilder.process(SSTableToCQL.java:694)
at com.scylladb.tools.SSTableToCQL$RowBuilder.access$200(SSTableToCQL.java:109)
at com.scylladb.tools.SSTableToCQL.process(SSTableToCQL.java:900)
at com.scylladb.tools.SSTableToCQL.stream(SSTableToCQL.java:931)
at com.scylladb.tools.BulkLoader.main(BulkLoader.java:1026)`

this exception comes from

**./java/com/scylladb/tools/SSTableToCQL.java**

error always references column name=h  which match the Clustering Key.

can this be due to data out of band ? or what can be triggering this exception?

Thanks

penberg:
@elcallio @haaawk, I think this is for you.

elcallio:
[...]

PR #368, opened 2023-12-19
build.xml: install and use java-11 when building

tchaikov:
before this change, we rely on JAVA_HOME to build for java11 target.
but user might set JAVA_HOME to a java-8, in this case the `javac`
from openjdk-8 would fail to compile, because it does not understand
the `--release` option:
```
    [javac] Compiling 1 source file to /home/piotrs/src/scylla2/tools/java/build/classes/main/META-INF/versions/11
    [javac] javac: invalid flag: --release
    [javac] Usage: javac <options> <source files>
    [javac] use -help for a list of possible options
```

in this change, we use the JAVA11_HOME environmental variable exposed by
`build_reloc.sh`, so that ant is able to build with the proper java even
if user sets JAVA_HOME to java-8.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -174,9 +174,6 @@
     Build instructions for development buils that only support Java 8:
     - requires JDK 8 in JAVA_HOME
     -->
-    <condition property="java.version.8">
-        <equals arg1="${ant.java.version}" arg2="1.8"/>
-    </condition>
     <condition property="java11-jvmargs" value="-Djdk.attach.allowAttachSelf=true --add-exports java.sql/java.sql=ALL-UNNAMED --add-exports java.base/jdk.internal.misc=ALL-UNNAMED --add-opens java.base/jdk.internal.module=ALL-UNNAMED --add-exports java.base/jdk.internal.ref=ALL-UNNAMED --add-exports java.base/sun.nio.ch=ALL-UNNAMED --add-exports java.management.rmi/com.sun.jmx.remote.internal.rmi=ALL-UNNAMED --add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED --add-exports java.rmi/sun.rmi.server=ALL-UNNAMED --add-opens jdk.management/com.sun.management.internal=ALL-UNNAMED">
         <not>
             <equals arg1="${ant.java.version}" arg2="1.8"/>
@@ -792,35 +789,7 @@
     <target name="build" depends="resolver-retrieve-build,build-project" description="Compile Cassandra classes"/>
     <target name="codecoverage" depends="jacoco-run,jacoco-report" description="Create code coverage report"/>
 
-    <target name="_build_java8_only" if="java.version.8">
-        <echo message="Compiling only for Java 8 ..."/>
-        <javac
-               debug="true" debuglevel="${debuglevel}" encoding="utf-8"
-               destdir="${build.classes.thrift}" includeantruntime="false" source="${source.version}" target="${target.version}"
-               memorymaximumsize="512M"
-               >
-            <src path="${interface.thrift.dir}/gen-java"/>
-            <classpath refid="cassandra.classpath"/>
-        </javac>
-        <javac fork="true"
-               debug="true" debuglevel="${debuglevel}" encoding="utf-8"
-               destdir="${build.classes.main}" includeantruntime="false" source="${source.version}" target="${target.version}"
-               memorymaximumsize="512M"
-               >
-            <src path="${build.src.java}"/>
-	    <!-- changed by calle: We don't care about the overhead of
-		 a reentrantlock in atomicbtreepartition, so always
-		 build againt "java 11", aka don't use forbidden api:s
-		 -->
-            <src path="${build.src.java11}"/>
-            <src path="${build.src.gen-java}"/>
-            <compilerarg value="-XDignore.symbol.file"/>
-            <classpath>
-                <path refid="cassandra.classpath"/>
-            </classpath>
-        </javac>
-    </target>
-    <target name="_build_multi_java" unless="java.version.8">
+    <target name="_build_multi_java">
         <!-- Note: we cannot use javac's 'release' option, as that does not allow accessing sun.misc.Unsafe nor
         Nashorn's ClassFilter class as any javac modules option is invalid for relase 8. -->
         <echo message="Compiling for Java 8 (using ${env.JAVA8_HOME}/bin/javac) ..."/>
@@ -856,7 +825,6 @@
             name="build-project">
         <echo message="${ant.project.name}: ${ant.file}"/>
         <!-- Order matters! -->
-        <antcall target="_build_java8_only"/>
         <antcall target="_build_multi_java"/>
         <antcall target="createVersionPropFile"/>
         <copy todir="${build.classes.main}">
@@ -1073,7 +1041,6 @@
 
     <!-- creates release tarballs -->
     <target name="artifacts" depends="jar,gen-doc,build-test,stress-build-test,sources-jar,javadoc-jar"
-            unless="java.version.8"
             description="Create Cassandra release artifacts">
       <mkdir dir="${dist.dir}"/>
       <!-- fix the control linefeed so that builds on windows works on linux -->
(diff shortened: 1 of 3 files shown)


tchaikov:
since we have FTBFS when developer happens to point `JAVA_HOME` to a java-1.8 installation. it'd be nice to fix this in the building system. so, this change resurrects #345.

tchaikov:
once this change is merged, and scylla picks up the submodule change, i will update https://github.com/scylladb/scylladb/blob/master/tools/toolchain/Dockerfile#L16 to remove 
```
ENV JAVA8_HOME=/usr/lib/jvm/java-1.8.0-openjdk
```

elcallio (approved):
I'm ok with this, but if we finally remove the java 8 compilation, why not just call JAVA11_HOME JAVA_HOME and just verify we have the appropriate level javac in build.xml?

nyh:
To be honest, I don't understand why you need to fix anything in this project, which is mostly a fork of Cassandra and could be built like Cassandra.

Personally, to compile scylla-tools-java I use the command:

```
ant realclean
JAVA_HOME=/usr/lib/jvm/java-11 JAVA8_HOME=/usr/lib/jvm/java-1.8.0 ant jar
```

I have a README to remind me of that.
You don't **have** to change the build.xml to explicitly override JAVA_HOME and JAVA8_HOME.

To compile more recent Cassandra, I need to use an even more elaborate command to have it avoid Java 17, but it also works:

```
JAVA_HOME=/usr/lib/jvm/java-11 JRE_HOME=/usr/lib/jvm/java-11/jre PATH=$JAVA_HOME:$JRE_HOME/bin:$PATH CASSANDRA_USE_JDK11=true ant -Duse.jdk11=true
```

What worries me is that I don't know if your changes, while improving the situation on your build environment, break it for other people with different build environments you didn't consider (I'm not saying it will - I'm just saying it worries me).

tchaikov:
> I'm ok with this, but if we finally remove the java 8 compilation, why not just call JAVA11_HOME JAVA_HOME and just verify we have the appropriate level javac in build.xml? 

Yeah. The original goal was to setup the right path for JAVA_HOME, but when I was reading your comment, I thought we could go further. IMHO, JAVA11_HOME is more explicit than JAVA_HOME. But I don't feel strong either way. I am not sure how to verify this in build.xml though. Just wanted to address the pain of FTBFS. The cleanup was but a bonus.

nyh:
> wanted to address the pain of those who fail to build this project. A README in one's $HOME directory or memory does not address this for developers who happen to run into this problem. I think the code is more reproducible than document. As always, i believe that developers' time is precious.

I agree with that, just make sure that after your changes it actually works for everyone, not just for you :-)
Since I already figured out how to build the existing project, I'm worried I'll need to learn new tricks to build it with your changes, if it doesn't just work out of the box:

Please correct me if I'm wrong, your patch doesn't just make "ant" work automatically - I still need to set JAVA11_HOME instead of the JAVA_HOME I had to set before. Right? If this is the case, I am not sure how it addresses the pain of people who build this project... Or maybe I'm missing something and just plain "ant" should work after your patch?

tchaikov:
> > wanted to address the pain of those who fail to build this project. A README in one's $HOME directory or memory does not address this for developers who happen to run into this problem. I think the code is more reproducible than document. As always, i believe that developers' time is precious.
> 
> I agree with that, just make sure that after your changes it actually works for everyone, not just for you :-) Since I already figured out how to build the existing project, I'm worried I'll need to learn new tricks to build it with your changes, if it doesn't just work out of the box:

no. it's not a breaking change. as i think it is not.

> 
> Please correct me if I'm wrong, your patch doesn't just make "ant" work automatically - I still need to set JAVA11_HOME instead of the JAVA_HOME I had to set before. Right? If this is the case, I am not sure how it addresses the pain of people who build this project... Or maybe I'm missing something and just plain "ant" should work after your patch?

the goal is to make the following workflow more smooth:

1. git clone scylladb and its submodules
2. `install-dependencies.sh`
3. use `./configure.py` to setup the building system
4. `ninja build/release/scylla`

one does should not need to setup `JAVA_HOME` or `JAVA11_HOME`. because the `build.ninja` generated by `configure.py` calls into `tools/java/build_reloc.sh`, which sets this up this for us, and `ant` picks the environmental variable up.

nyh:
As far as I know (please correct me if I'm wrong), `ninja build/release/scylla` doesn't call tools/java/build_reloc.sh. We have other scripts that do this, but not `ninja build/release/scylla`.

Anyway, if you don't care what "cd tools/java; ant" does, why did you want to modify build.xml at all? You could have done anything you wanted in the scripts you mentioned - including "unset JAVA_HOME" if you wanted to.

Anyway, I'm not against this patch - all that I'm saying, I guess, is that I don't understand it :-)

tchaikov:
> As far as I know (please correct me if I'm wrong), `ninja build/release/scylla` doesn't call tools/java/build_reloc.sh. We have other scripts that do this, but not `ninja build/release/scylla`.

sorry. i failed to copy the command line and instead typed the one i was using. it should be 

```bash
ninja  -j14
```

> 
> Anyway, if you don't care what "cd tools/java; ant" does, why did you want to modify build.xml at all? You could have done anything you wanted in the scripts you mentioned - including "unset JAVA_HOME" if you wanted to.
> 

agreed. a minimal change does not need to touch `build.xml`. that'd be more or less what #345 does. but the more i read the comments in that PR the more i was tempted to go further: if we don't need this and that, why would we need to install and build foo and bar, and how about ... hence these changes.

> Anyway, I'm not against this patch - all that I'm saying, I guess, is that I don't understand it :-)

tchaikov:
v2:

- s/JAVA11_HOME/JAVA_HOME/ to preserve the existing behavior.

elcallio:
LGTM

tchaikov:
v3:

- rebase against master:HEAD

tchaikov:
this fails the build. see https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/1961/consoleText

```
[2023-12-22T06:59:52.909Z] [151/1070] cd tools/java && ./reloc/build_reloc.sh --version $(<../../build/SCYLLA-PRODUCT-FILE)-$(sed 's/-/~/' <../../build/SCYLLA-VERSION-FILE)-$(<../../build/SCYLLA-RELEASE-FILE) --nodeps 
[2023-12-22T06:59:52.909Z] FAILED: tools/java/build/scylla-tools-5.5.0~dev-0.20231222.7b1a0af13993.noarch.tar.gz 
[2023-12-22T06:59:52.909Z] cd tools/java && ./reloc/build_reloc.sh --version $(<../../build/SCYLLA-PRODUCT-FILE)-$(sed 's/-/~/' <../../build/SCYLLA-VERSION-FILE)-$(<../../build/SCYLLA-RELEASE-FILE) --nodeps 
[2023-12-22T06:59:52.909Z] unable to find jdk-11
```

we need to install `openjdk-11-jdk-headless` to fix it.

tchaikov:
v4:

- expose `JAVA_HOME` if java-17 is found. this addresses the FTBFS found in https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/1961/consoleText. 

the build and test run can be found at https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/1963/console

tchaikov:
> byo

https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/2451/consoleText . still the same.

tchaikov:
anyway, i think there is no need to get this in, as

- we are using the native nodetool nowadays, and
- ppl are using dbuild for building scylla

in short, for the same reason put at #207. i am not going to continue pushing forwards this PR anymore.

tchaikov:
@scylladb/scylla-maint hello maintainers, could you help install `openjdk-11-jdk-headless` to the dbuild image? 
we are seeing build failures in our CI due to the `--release` problem:

- https://jenkins.scylladb.com/job/scylla-master/job/scylla-ci/11762/console
- https://jenkins.scylladb.com/job/scylla-master/job/next/8228/

would be nice if we can fix it once and for all.

avikivity:
I agree with the change, but I don't want to bring it in for a toolchain update.

Also, do we need -devel, or is the jdk enough?

avikivity:
I see -devel and openjdk are interdependent.

avikivity:
Turns out adding java-11-openjdk wasn't enough

tchaikov:
> Turns out adding java-11-openjdk wasn't enough

@avikivity adding java-11-openjdk is not enough, but `openjdk-11-jdk-headless` should suffice. the former run a java application, the later build it.

avikivity:
openjdk-headless is a subset of openjdk-devel, no?

avikivity:
It doesn't contain javac for instance. We need -devel.

avikivity:
It doesn't work installing random stuff, and does work with this series. My conclusion is that something changed in how Fedora selects Java, and this series fixes it (although: it works on my workstation without the series).

PR #369, opened 2023-12-20
treewide: update "guava" package

tchaikov:
update "guava" package from 18.0 to 32.1.3.

Update the version of guava dependency to 31.1.3-jre. Before the change, security scanners (such as Trivy) reported that `guava` used in the project was vulnerable to CVE-2018-10237 and CVE-2023-2976 (both "MEDIUM" severify and CVE-2020-8908 as "LOW" severity (both "HIGH" severity).

Those issues were fixed in guava 31.1.3 and after this commit the security scanner doesn't report any problems related to this dependency.

because guava 31 introduced quite a few non-backward compatible changes, we have to address them on a case-by-case basis.

previous this change was reverted in
3963c3abf71a6df310ca8f3849e4cf8562469666 because it missed the change to address the incompatible API changes.

Fixes: https://github.com/scylladb/scylla-tools-java/issues/365

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -363,11 +363,10 @@
           <dependency groupId="org.xerial.snappy" artifactId="snappy-java" version="1.1.10.4"/>
           <dependency groupId="net.jpountz.lz4" artifactId="lz4" version="1.3.0"/>
           <dependency groupId="com.ning" artifactId="compress-lzf" version="0.8.4"/>
-          <dependency groupId="com.google.guava" artifactId="guava" version="18.0">
+          <dependency groupId="com.google.guava" artifactId="guava" version="32.1.3-jre">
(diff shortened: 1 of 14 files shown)


yaronkaikov (approved):
LGTM. Thanks @tchaikov for your help with that

yaronkaikov:
@scylladb/scylla-maint Can you please merge this and update the submodule ?

roydahan:
@yaronkaikov so this one is in?
I thought you were saying it will break next if it gets in.

yaronkaikov:
It merged here, but the submodule is not updated in Scylla core until we fix https://github.com/scylladb/java-driver/pull/267

roydahan:
@avelanarius FYI

Issue #371, opened 2024-01-04
scylla-tools-java deps - update JNA (if we need to)


mykaul:
https://github.com/scylladb/scylla-tools-java/blob/95b3c9dd4095c40729605d650dc1011e192fe93c/build.xml#L429 is using a too old version. Cassandra has moved to a much newer version (see https://github.com/apache/cassandra/commit/a8b1fb6ae6e934f4ba0a07deb03b6257c7dcb383 ).

mykaul:
The 'if we need to' is that it's unclear if we need it for cassandara-stress or the tools we do use in this repo in Java...

mykaul:
@roydahan - per our discussion today.

mykaul:
@scylladb/scylla-tools-maint  - please backport to both 5.4 and 5.2.

denesb:
Backported to 5.4 and 5.2.

PR #376, opened 2024-01-11
Update JNA dependency to 5.14.0

mykaul:
Just compile-tested!

Fixes: https://github.com/scylladb/scylla-tools-java/issues/371


yaronkaikov:
you can run locally `./reloc/build_reloc.sh` and see if it's compiling

mykaul:
Yes, indeed!

PR #380, opened 2024-02-11
bin/nodetool-wrapper: override the -p option with the one set by -Dcom.scylladb.apiPort

tchaikov:
"scylla nodetool" connects to the port of the API service, while
Cassandra nodetool connects to the port of the JMX service. and
nodetool-wrapper tries to be a broker to delegate to the latter
if the "scylla nodetool" cannot handle the specified command. but, if
the command line has "-p 7199" instead of "-p 10000" specified,
and the command line is able to be handled by "scylla nodetool",
nodetool-wrapper would be instructing "scylla nodetool" to connect
to JMX server instead of the API service. that's why we'd have
following test failure when we tries to enable "scylla nodetool"
to handle more commands in the test:

```
ccmlib.node.ToolError: Subprocess /jenkins/workspace/scylla-master/gating-dtest-release/scylla/.ccm/scylla-repository/17238/share/cassandra/bin/nodetool -h 127.0.18.1 -p 7199 -Dcom.sun.jndi.rmiURLParsing=legacy flush exited with non-zero status; exit status: 2;
stderr: error running operation: std::runtime_error (Invalid response)
```

in this change, we let the wrapper to mutate the "-p" or "--port" option
in the command line options with the one specified by
"-Dcom.scylladb.apiPort=<num>", which is not recognized by
Cassandra nodetool, but it is set by the ccmlib, when the test is
performed against a scylla server. so, both "scylla nodetool"
can connect to the API service when it is called by the tests in
dtest.


bin/nodetool-wrapper
```diff
@@ -22,21 +22,26 @@ fi
...
-for opt in "$@"; do
-    # if "--help" is already in the command line, no need to added it for
-    # checking scylla-nodetool implements the given command.
-    if [ "$opt" = "$help_opt" ]; then
-        help_opt=""
-        break
-    fi
-done
 
+help_opt=$(get_help_opt $@)
```
avikivity, line 85: "$@" should be quoted, or you introduce more word breaks when you pass it.

avikivity: I also don't see how it helps readability

bin/nodetool-wrapper
```diff
@@ -35,6 +35,47 @@ function get_help_opt() {
...
+        case "$1" in
+            -Dcom.scylladb.apiPort=*)
+                # swallow it
+                shift
+                ;;
+            -p|--port)
+                # override the argument
+                new_args+=("$1")
+                new_args+=("$api_port")
+                shift 2
```
avikivity, line 68: Do both the original and new nodetool support a -p option? That's wrong since they refer to different ports.

Better have a new option for the REST port.

tchaikov: > Do both the original and new nodetool support a -p option? That's wrong since they refer to different ports.

yes, they both support 

- `-p <port>`, `--port=<port>` and `--port <port>`, 
- `-h <host>`, `--host=<host>` and `--host <host>`

> 
> Better have a new option for the REST port.

IMHO, from the `scylla nodetool` 's perspective, `-p` and `--port` is a natural choice. as it just talks to the API port. as an alternative, probably we could add a new option to the nodetool-wrapper? but that's exactly what `-Dcom.scylladb.apiPort` does. i didn't add a proper option because i didn't wanted

1. filter the new option for the cassandra nodetool tool
2. translate it to `-p` for "scylla nodetool"
3. remove the `-p` for "scylla nodetool"

with `-Dcom.scylladb.apiPort`, i just need to 

1. override it to `-p` for "scylla nodetool"

bin/nodetool-wrapper
tchaikov: a bulk of code list in the body *could* make it difficult to parse. the longer the code list is the more difficult to maintain, we probably don't have a shell script which is of thousands of lines. but if we have, to figure out what it does i would have to read it line by line, to figure the data dependencies. before the `nodetool-wrapper` grows into the expanded form, i intend to make it more structured. so one does not have to dive into individual functions to understand the structure of the script, or to even hack it.

tchaikov: > "$@" should be quoted, or you introduce more word breaks when you pass it.

thank you for pointing this out. the quotes in shell is always a nightmare. indeed. fixed.

tchaikov:
v2:

- quote `$@`

bin/nodetool-wrapper
denesb: I would prefer if any hacks around the transition from the legacy nodetool to the native nodetool, would be concentrated in this file as much as possible. So that when we are done with the transition we can just drop this file and drop all the hacks together with it.

Thinking ahead, having a single `-p|--port` option for the native nodetool is preferable, then having an `--api-port` and a dead `-p|--port`, which is silently ignored because it used to refer to the now defunct JMX port.

IMO we should strive to have a clear interface long-term and tolerate some confusion during the transition if we have to.

denesb: (the above is to say I agree with the current approach and I prefer it instead of adding a new port option)

tchaikov:
v3:

- i forgot to pass `$api_port` to `override_port_option`, fixed now.

bin/nodetool-wrapper
avikivity: It looks cleaner but doesn't work. If you had a script that called nodetool with the jmx port, it wouldn't work with native nodetool, and vice versa.

tchaikov: @denesb hi Botond, i am convinced by Avi now. we might need to add a `--api-port` to *this script* before fully migrating to "scylla nodetool". by then we can drop this option, well, not exactly, we would be dropping this script. and everything would just work fine.

bin/nodetool-wrapper
nyh: You just need to know that `"$@"`, with the quotes, is a shell idiom. There is no place where `$@` without quotes is correct.

bin/nodetool-wrapper
nyh: I was actually convinced by Botond ;-) A user reads in some document that you need to pass the port with "-p", and he doesn't care what this port means, whether it is JMX or REST API.
Using this option would be rare anyway, but if a user does need to choose a port explicitly, no need to invent two differently-named options to do it.

denesb: If we add `--api-port` to this script, we also have to add it to `scylla-nodetool` and also have to teach it to expect both to be set, and which one to choose if so. We can do that to ease the transition.

avikivity: We should change scylla-nodetool to reject --port.

This is similar to accepting a measurement without units. The tools don't need "a port". They need a port that implements a specific protocol, but each wants a different protocol. They're not compatible in that respect, and we shouldn't pretend they are.

Yes, it's ugly if the way to specify scylla-nodetool's port isn't --port, but that's life.

denesb: By the same vein, we should reject `-h|--host`. And what about the other legacy JMX related parameters which we silently accept (and ignore)? I think adding two more paremeters to the already longish list of parameters we reinterpret/ignore is not breaking anything here.

Our compatibility is pretty shallow after all and so far our policy has been to cause the least surprise possible while not standing in the way. I think rejecting `--host` and `--port` would definitely qualify as "standing in the way".

nyh: I still think that whatever scripts do with these options is not relevant and not a problem. The only thing I'm thinking about is **documentation** - people will look on Google for nodetool documentation, and hope that whatever examples they see there, just works. If the example tells you "you can use -p to specify a different port", you try that.
But to be honest, I don't think it's very important. If "-p" doesn't work, the user will just use "--help" and expect to see the correct list of options. Just please make sure that "--help" isn't forwarded to the other nodetool :-)

denesb:
@tchaikov I think we don't need this PR anymore, with v2 of https://github.com/scylladb/scylladb/pull/17168.
The latter adds native support for `-Dcom.scylladb.apiPort` in `scylla-nodetool`, so we can just let the option pass-through unchanged.

tchaikov:
ack. closing this change.

PR #384, opened 2024-03-22
dist: drop Java tools and change packaging just for cassandra-stress

### LINKED ISSUE #370: Separate cassandra-stress to its own build, package, container
As we depart from Java based tooling, slowly but surely, we do not need to continue and have this Java overhead with us everywhere (specific example - no use of cassandra-stress in a Dockerized Scylla - so why have Java there?).

We should split it off and *if* we wish to, provide it separately.

CC @yaronkaikov

syuu1228:
Since Java based tools are deprecated and now we only need cassandra-stress, change package name to "scylla-cassandra-stress" and drop deprecated tools from the package.

Closes #370


dist/debian/control.template
```diff
@@ -7,16 +7,35 @@ Standards-Version: 3.9.5
 Rules-Requires-Root: no
 
 Package: %{product}-tools
+Depends: %{product}-cassandra-stress, ${misc:Depends}
 Architecture: all
-Depends: %{product}-conf, %{product}-tools-core, ${misc:Depends}
+Priority: optional
```
tchaikov, line 12: since you are specifying the "Priority" of this package. i'd suggest either do the same to `${product}-cassandra-stress`, or just drop both of them. because, according to [Debian Policy Manual](https://www.debian.org/doc/debian-policy/ch-archive.html#priorities), 

> extra
>    This priority is deprecated. Use the optional priority instead. This priority should be treated as equivalent to optional.

dist/debian/control.template
```diff
@@ -7,16 +7,35 @@ Standards-Version: 3.9.5
...
+Architecture: all
+Priority: optional
+Section: oldlibs
+Description: transitional package
+ This is a transitional package. It can safely be removed.
+
+Package: %{product}-cassandra-stress
+Architecture: all
+Depends: %{product}-conf, %{product}-cassandra-stress-core, ${misc:Depends}
 Recommends: ntp | time-daemon
```
tchaikov, line 28: `ntp` and `time-daemon` were added in https://issues.apache.org/jira/browse/CASSANDRA-4606 and https://issues.apache.org/jira/browse/CASSANDRA-5816 respectively. they intended to run a cassandra *cluster* . but cassandra-stress as a client-side tool, it does not rely on a ntp or chrony service to operate. so i'd suggest take this opportunity to drop them.

tchaikov (approved):
lgtm. just a couple nits. but they are not blockers.

denesb:
I think https://github.com/scylladb/scylla-ccm/pull/565 has to go in, before we can merge this.

dist/debian/control.template
syuu1228: Added `Priority: optional` to %{product}-cassandra-stress and %{product}-cassandra-stress.

dist/debian/control.template
syuu1228: Dropped.

yaronkaikov:
> I think [scylladb/scylla-ccm#565](https://github.com/scylladb/scylla-ccm/pull/565) has to go in, before we can merge this.

@denesb It was merged yesterday, can we merge this one?

denesb:
> > I think [scylladb/scylla-ccm#565](https://github.com/scylladb/scylla-ccm/pull/565) has to go in, before we can merge this.
> 
> @denesb It was merged yesterday, can we merge this one?

Not yet. The above PR was reduced in scope to just prepare ccm for the post JMX world. Throwing out all of Java on on go proved too big of a bite, to chew at once.

I just realized we also didn't yet finish https://github.com/scylladb/scylladb/issues/14856, notably, there are still dtests around testing the java tools. These need to be either:
* Dropped altogether, if we decided we don't want/need a native replacement for the tested tool (e.g. `sstableutil`).
* Rewrite the test to test the native variant. We are still lacking native variant to some of these (small ones but still).

syuu1228:
Rebased, since it was conflicted.

syuu1228:
I need to modify Obsoletes/Breaks/Replaces on this patch, since I supposed next release version is 5.5 so I wrote something like `Replaces: %{product}-tools (<< 5.5)` in this patch, but now we already released 6.0.

syuu1228:
Replaced `<< 5.5` with `<< 6.1`.

roydahan:
I think this one is not needed and this task https://github.com/scylladb/scylla-tools-java/issues/370 misled you.

We can/should keep the packaging as "scylla-tools-java" and c-s will be part of it.
There is no need to separate it now after we separated it from Scylla.

So, the packaging can remain as scylla-tools-java but just not a dependency of scylladb.

yaronkaikov:
@syuu1228 I guess we can continue with that now, right?

syuu1228:
> I think this one is not needed and this task #370 misled you.
> 
> We can/should keep the packaging as "scylla-tools-java" and c-s will be part of it. There is no need to separate it now after we separated it from Scylla.
> 
> So, the packaging can remain as scylla-tools-java but just not a dependency of scylladb.

Okay, closing this PR.

PR #386, opened 2024-04-17
install.sh: don't install nodetool

denesb:
Nodetool is now supplied by the scylla core package, the Java tools package shouldn't install it anymore.
For now, leave it as a backup nodetool in /opt/scylladb/share/cassandra/bin, but don't symlink it into /usr/bin.

After merged, the submodule update for this PR should be incorporated into https://github.com/scylladb/scylladb/pull/18255.


mykaul:
When we upgrade the package, will it remove the now defunct Java based nodetool? In the RPM post install script for example?

denesb:
> When we upgrade the package, will it remove the now defunct Java based nodetool? In the RPM post install script for example?

No, and we should. I don't know how to do it though. Will need some help on that.

denesb:
BTW, the `nodetool-wrapper` removed by this PR didn't make it into any release yet, so no need to add rules to remove it.
We should clear `nodetool` binary from `/opt/shared/cassandra/bin`.

dist/debian/debian/scylla-tools.install
```diff
@@ -1,15 +1,10 @@
...
-opt/scylladb/share/cassandra/bin/nodetool-java
 opt/scylladb/share/cassandra/bin/sstableloader
 opt/scylladb/share/cassandra/bin/cassandra-stress
 opt/scylladb/share/cassandra/bin/cassandra-stressd
 opt/scylladb/share/cassandra/bin/sstabledump
 opt/scylladb/share/cassandra/bin/sstablelevelreset
 opt/scylladb/share/cassandra/bin/sstablemetadata
 opt/scylladb/share/cassandra/bin/sstablerepairedset
-usr/bin/nodetool
-usr/bin/nodetool-java
```
tchaikov, line 12: in debian, we have a dedicated process for handling the package transition. see https://www.debian.org/doc/debian-policy/ch-relationships.html#s-replaces . in this case, the files under the name of `nodetool` are moved from `scylla-tools` into `scylla-server`. this case is the 9th case documented by https://wiki.debian.org/PackageTransition .

> Reorg: A and B existed; move some files from A to B; new A does not require new B 

so we need to apply
```
Flags for new A package  flags for new B package
 Breaks: B (<<2)         Breaks: A (<<2)
                         Replaces: A (<<2) 
```

where A is `scylla-tools`, B is `scylla-server`. and assuming we are doing this transition from 5.x to 6.0.

in other words, we need to have following dependencies in 6.0's packaging:

1. for `scylla-tools` , it `Breaks scylla-server (<<6.0)`
2. for `scylla-server`, it `Breaks scylla-tools (<<6.0)` and it `Replaces: scylla-tools (<<6.0)`

denesb: Can we use `replaces` and `breaks`, even if we are not replacing the entire content of the package yet?

tchaikov: @denesb yes we can, and we should. `scylla-tools` is not a meta package, this granularity is the finest we can achieve.

denesb: Ok, I will try to implement this.

denesb:
v2:
* Keep installing nodetool to `/opt/scylladb/share/cassandra/bin/nodetool`, but don't create a thunk/symlink in `/usr/bin` (in root install).

@tchaikov after thinking about it, I don't think I need to do any transition work in the packaging of `scylla-tools`. Instead all the required changes have to go into the packaging of `scylla-server`, as it is the one which takes over files from this package, the reverse is not true.

tchaikov:
> v2:
> 
> * Keep installing nodetool to `/opt/scylladb/share/cassandra/bin/nodetool`, but don't create a thunk/symlink in `/usr/bin` (in root install).
> 
> 
> 
> @tchaikov after thinking about it, I don't think I need to do any transition work in the packaging of `scylla-tools`. Instead all the required changes have to go into the packaging of `scylla-server`, as it is the one which takes over files from this package, the reverse is not true.

agreed, I was commenting on the wrong PR. All the comments should apply to the scylladb PR.

tchaikov:
> > v2:
> > 
> > * Keep installing nodetool to `/opt/scylladb/share/cassandra/bin/nodetool`, but don't create a thunk/symlink in `/usr/bin` (in root install).
> > 
> > @tchaikov after thinking about it, I don't think I need to do any transition work in the packaging of `scylla-tools`. Instead all the required changes have to go into the packaging of `scylla-server`, as it is the one which takes over files from this package, the reverse is not true.
> 
> agreed, I was commenting on the wrong PR. All the comments should apply to the scylladb PR.

> for scylla-tools , it Breaks scylla-server (<<6.0)

i was wrong. the above comment still applies.

denesb:
> > > 
> > 
> > 
> > agreed, I was commenting on the wrong PR. All the comments should apply to the scylladb PR.
> 
> > for scylla-tools , it Breaks scylla-server (<<6.0)
> 
> i was wrong. the above comment still applies.

I don't understand. Why is this package breaking scylla-server? Older versions of this package break the new version of scylla-server, but the new one doesn't, they have no common files.

tchaikov:
sorry for the confusion, i meant: the new scylla-tools should not coexist with old scylla-server, that's why the new `scylla-tools` "breaks" the old `scylla-server`.

denesb:
> sorry for the confusion, i meant: the new scylla-tools should not coexist with old scylla-server, that's why the new `scylla-tools` "breaks" the old `scylla-server`.

Ok, makes sense.

tchaikov (approved):
lgtm

denesb:
@scylladb/scylla-maint please merge. Both this and its sister PR https://github.com/scylladb/scylladb/pull/18255 are ready for merge.

denesb:
(remember to not create a submodule update after merging, the submodule update has to go into https://github.com/scylladb/scylladb/pull/18255, so it is enough to just merge this to scylla-tools-java:master).

PR #389, opened 2024-04-22
install-dependencies.sh: s/python/python3/

tchaikov:
python is EOL. and the scripts in this repo actually are using python3 instead python:

```console
$ find . -name '*.py' -exec grep '#!' {} \;
```
also, "python" as a package name is not available in debian and and some of its derivative distros, see
https://packages.debian.org/search?keywords=python&searchon=names&suite=stable&section=all and
https://packages.ubuntu.com/search?suite=jammy&section=all&arch=any&keywords=python3&searchon=names

but python3, or better off python3-minimal is.
see
https://packages.debian.org/search?keywords=python3-minimal&searchon=names&suite=stable&section=all and
https://packages.ubuntu.com/search?suite=jammy&section=all&arch=any&keywords=python3-minimal&searchon=names

ubuntu/jammy is picked, because it is currently the ubuntu-latest os provided by github workflow runners.

so, in this change, let's replace "python" with "python3", so that we can run `install-dependencies.sh` without error like:

```
Package python is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  2to3 python2-minimal python2 dh-python python-is-python3

E: Package 'python' has no installation candidate
```

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -20,7 +20,7 @@
 . /etc/os-release
 
 if [ "$ID" = "ubuntu" ] || [ "$ID" = "debian" ]; then
-    apt -y install openjdk-8-jdk-headless ant ant-optional python
+    apt -y install openjdk-8-jdk-headless ant ant-optional python3
 elif [ "$ID" = "fedora" ] || [ "$ID" = "centos" ]; then
-    dnf install -y --setopt=install_weak_deps=False ant java-1.8.0-openjdk-devel python ant-junit fakeroot
+    dnf install -y --setopt=install_weak_deps=False ant java-1.8.0-openjdk-devel python3 ant-junit fakeroot
 fi


tchaikov:
@scylladb/scylla-maint hi maintainers, could you help review this change? i am trying to add a workflow to scylladb using github action, which failed like https://github.com/tchaikov/scylladb/actions/runs/8778841191/job/24085931123 .

would be great if we can use github's ubuntu-latest runner to perform some static analysis.

denesb:
I wonder if we need to regenerate the toolchain image for this PR. Does this change make a difference on fedora?
I guess we should regenerate it anyway just to really test the change.

tchaikov:
> I wonder if we need to regenerate the toolchain image for this PR. 

yeah, to be on the safe side, we should.

> Does this change make a difference on fedora?

yes. as i am changing the feodra branch as well. but i don't have to. i did it just for the same of completeness. what i care is but the ubuntu runner.

> I guess we should regenerate it anyway just to really test the change.

if it's complicated. or does not worth the efforts. i have some workarounds:

1. only change the ubuntu/debian branch
2. edit the `install-dependencies.sh` in scylladb in the workflow, so that it does not call these java tools' `install-dependencies.sh` (actually i am working in this direction)
3. wait until we drop this submodule (it will take longer, as we are still packaging nodetool, and we will be using cassandra-stress, in a relatively long run)

denesb:
> if it's complicated. or does not worth the efforts. i have some workarounds:

It is not complicated at all. I just execute a command and a few hours later we have the new toolchain. The hardest part is just remembering that one started a regenerate job half a day later.

tchaikov:
i see. thank you Botond!

denesb:
Actually, it is a bit more complicated because I just noticed that this is the java-tools submodule. So it needs some coordination.

mykaul:
Where do we use Python in scylla-tools-*java* ? (note that CQLSH has moved to its own repo long time ago)

tchaikov:
> Where do we use Python in scylla-tools-_java_ ? (note that CQLSH has moved to its own repo long time ago)

```console
$ cd ~/dev/scylladb/tools/java/
$  find . -name '*.py'
dist/debian/debian_files_gen.py
doc/scripts/convert_yaml_to_adoc.py
doc/scripts/gen-nodetool-docs.py
scripts/create-relocatable-package.py
```

@mykaul hi Yaniv, we are just using python to build the reloc package and debian package. for the sake of self-containness, i think it's fair to include `python` (should be python3!) in the `install-dependencies.sh` in this repo.

mykaul:
> @mykaul hi Yaniv, we are just using python to build the reloc package and debian package. for the sake of self-containness, i think it's fair to include `python` (should be python3!) in the `install-dependencies.sh` in this repo.

I don't think build-time deps should be included - especially not if it's 'just' for packaging. Moreover - two scripts are for docs only - which never/rarely change, one is for Debian package (which is just odd? Why don't we have the package metadata?) and the relocatable is the only one truly needed, no?

denesb:
Started regenerating the toolchain.

denesb:
> I don't think build-time deps should be included - especially not if it's 'just' for packaging. Moreover - two scripts are for docs only - which never/rarely change, one is for Debian package (which is just odd? Why don't we have the package metadata?) and the relocatable is the only one truly needed, no?

I think it is very much expected by anybody executing `./install-dependencies.sh` to get all dependencies, build-time or run-time. BTW scylladb.git also pulls in python3 already, so dropping it from here will not change anything.

mykaul:
> > I don't think build-time deps should be included - especially not if it's 'just' for packaging. Moreover - two scripts are for docs only - which never/rarely change, one is for Debian package (which is just odd? Why don't we have the package metadata?) and the relocatable is the only one truly needed, no?
> 
> I think it is very much expected by anybody executing `./install-dependencies.sh` to get all dependencies, build-time or run-time. BTW scylladb.git also pulls in python3 already, so dropping it from here will not change anything.

OK, was hoping to slowly reduce our deps all around. Specifically, get rid of Python in the (future) Scylla-only container.

denesb:
> Started regenerating the toolchain.

PR with regenerated toolchain: https://github.com/scylladb/scylladb/pull/18394

Once CI passes, I will merge this and rebase that PR to include the submodule update.

denesb:
submodule update PR: https://github.com/scylladb/scylladb/pull/18394

denesb:
Removed from master, because we currently cannot generate new toolchains, due some changes around the PGO compilers. @tchaikov please re-submit this PR and I will merge once toolchains can be generated again.

tchaikov:
@denesb thank you for your efforts. i recreated this change at #391

PR #395, opened 2024-06-10
build: don't add nonexistent directory 'lib' to relocatable packages

avikivity:
The call to reloc_add('lib') fails with Python 12 as there is no such directory. Somehow it worked in older versions.

We later add a file to that directory, but tar files support having a file in a directory not named earlier.

Fix by removing the call.


avikivity:
@denesb please review

denesb (approved):

denesb:
Submodule update: scylladb/scylladb@56ec839f41d137a39fe0a54ca760200674697c83.

PR #397, opened 2024-06-11
Revert "build: don't add nonexistent directory 'lib' to relocatable p…

tchaikov:
…ackages"

This reverts commit 82e3d0c6a894ac752a67d1342e8ca7e1192cde5f.

the "lib" directory is not empty, and it collects all the runtime dependencies for running the java-based nodetool. without them, we'd run into test failures when testing dtests using the java-based nodetool:

```
>           raise ToolError(cmd_args, rc, out, err)
E           ccmlib.node.ToolError: Subprocess ['stress', 'write', 'n=50K', '-rate', 'threads=4', '-node', '127.0.96.1'] exited with non-zero status; exit status: 1;
E           stderr: Exception in thread "main" java.lang.NoClassDefFoundError: io/netty/util/concurrent/FastThreadLocal
E           	at org.apache.cassandra.stress.Stress.main(Stress.java:59)
E           Caused by: java.lang.ClassNotFoundException: io.netty.util.concurrent.FastThreadLocal
E           	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
E           	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
E           	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:527)
E           	... 1 more
```


avikivity:
Wow, that was terrible. Why did I think it was empty?

PR #401, opened 2024-07-04
install.sh: introduce Java existance check

syuu1228:
Since we dropped Java check from unified installer and jmx, we need to do it on scylla-tools.

Related https://github.com/scylladb/scylladb/pull/17969


syuu1228:
This is part of https://github.com/scylladb/scylladb/pull/17969, need to merge first.

scripts/select-java
```diff
@@ -0,0 +1,30 @@
...
+            javaver=$($java -XshowSettings:properties -version 2>&1 | awk -F' = ' '/java.specification.version/ {print $NF}')
+            if [ "$javaver" = $expected_java_version ]; then
+                echo $java
+                return
+            fi
+        done
+    done
+}
+
+# So far, scylla-jmx only works with Java 11 and Java 8, we prefer the newer
```
tchaikov, line 19: might want to s/scylla-jmx/scylla-tools/ . probably we could be more specific though, like "tools like nodetool and cassandra-stress". but "scylla-tools" is the name used by https://github.com/scylladb/scylla-tools-java?tab=readme-ov-file#scylla-tools .

tchaikov (commented):
also, in scylla-jmx,  `select-java` actually serves for use cases

- if we are not installing the package, it selects the java executable at runtime, see https://github.com/scylladb/scylla-jmx/blob/3328a229610daac8198a4133b496a9d7cb1de469/scripts/scylla-jmx#L134 .
- if we are installing the package using `install.sh`, `install.sh` is responsible for detecting the runtime environment for the java applications. so it also checks for the available java runtimes.

in this change, we only address the second use case, where we are installing scylla-tools using `install.sh`, but what about the first use case? please note, the existing check is performed by https://github.com/scylladb/scylla-tools-java/blob/01ba3c196f2007dcd1bed3755bd5d392be4f499a/dist/common/cassandra.in.sh#L65 , which just **check** if the `java` fulfills the needs, it does not pick a java runtime. so if user actually has openjdk-11 installed, but he/she doe not point `/usr/bin/java` to it, or export `$JAVA` so that it points to the qualified java runtime, `cassandra.in.sh` fails.

syuu1228:
> also, in scylla-jmx, `select-java` actually serves for use cases
> 
> * if we are not installing the package, it selects the java executable at runtime, see https://github.com/scylladb/scylla-jmx/blob/3328a229610daac8198a4133b496a9d7cb1de469/scripts/scylla-jmx#L134 .
> * if we are installing the package using `install.sh`, `install.sh` is responsible for detecting the runtime environment for the java applications. so it also checks for the available java runtimes.
> 
> in this change, we only address the second use case, where we are installing scylla-tools using `install.sh`, but what about the first use case? please note, the existing check is performed by
> 
> https://github.com/scylladb/scylla-tools-java/blob/01ba3c196f2007dcd1bed3755bd5d392be4f499a/dist/common/cassandra.in.sh#L65
> 
> , which just **check** if the `java` fulfills the needs, it does not pick a java runtime. so if user actually has openjdk-11 installed, but he/she doe not point `/usr/bin/java` to it, or export `$JAVA` so that it points to one of the supported java runtime, `cassandra.in.sh` fails.

We need some code to detect Java on scylla-tools's install.sh, since we are dropping it from unified installer.
But, as you descripbed, Java detection method between select-java and cassandra.in.sh are different, I think we should not use select-java here.
So I pushed new code to detect Java just like cassandra.in.sh, and dropped select-java.

install.sh
```diff
@@ -68,6 +74,24 @@ while [ $# -gt 0 ]; do
...
+                JAVA="$java"
+                break
+            fi
+        done
+    else
+        JAVA=java
+    fi
+
+    if ! builtin command -v $JAVA > /dev/null; then
+        echo "Please install openjdk-11 before running install.sh."
```
tchaikov, line 90: what if user installs, for instance, openjdk-21 and have it in `$PATH`? and we are not checking for openjdk-11, but a random `java` executable in the `$PATH`. so even if this test passes, it does not imply that we have openjdk-11 around.

PR #402, opened 2024-07-16
cassandra-stress: Make default repl. strategy NetworkTopologyStrategy

### LINKED ISSUE #400: stress should use the NetworkTopologyStrategy by default, not SimpleStrategy
Curreently stress uses `SimpleStrategy` as the default when creating the keyspace.
This strategy is deprecated in scylla and it inhibits tablets, which require NetworkTopologyStrategy,
so let's use the latter instead.

elcallio:
Fixes #400

Switch default strategy to scylla preferred default.

### DIFF
--- a/tools/stress/src/org/apache/cassandra/stress/settings/OptionReplication.java
+++ b/tools/stress/src/org/apache/cassandra/stress/settings/OptionReplication.java
@@ -34,8 +34,9 @@ import org.apache.cassandra.locator.AbstractReplicationStrategy;
  */
 class OptionReplication extends OptionMulti
 {
+    private static final String DEFAULT_STRATEGY = "org.apache.cassandra.locator.NetworkTopologyStrategy";
 
-    private final OptionSimple strategy = new OptionSimple("strategy=", new StrategyAdapter(), "org.apache.cassandra.locator.SimpleStrategy", "The replication strategy to use", false);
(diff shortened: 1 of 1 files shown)


bhalevy (approved):

bhalevy:
@scylladb/scylla-tools-maint please merge

denesb:
https://github.com/scylladb/scylladb/pull/19818

bhalevy:
@dani-tweig can you please add backport labels for 6.1 and 6.2 (and we probably also need backport labels for enterprise branch in scylla-enterprise-tools-java)?

This needs backport to 6.1 and 6.0.
Cc @denesb

denesb:
@bhalevy why should we backport this? I need a scylla issue with a problem the this PR and the backport would fix.

bhalevy:
> @bhalevy why should we backport this? I need a scylla issue with a problem the this PR and the backport would fix.

The issue is https://github.com/scylladb/scylla-tools-java/issues/400.

We want to use NetworkTopologyStrategy by default for 2 reasons:
1. To enable tablets for cassandra-stress when they are enabled by default - this is the motivation for the backport
2. SimpleStrategy was deprecated a LONG time ago, and it's about time we stop using it by default in all of our tools (no other reason is really required) - this does not require backport.

denesb:
Backports:
* 6.1: https://github.com/scylladb/scylladb/pull/20199
* 6.0: https://github.com/scylladb/scylladb/pull/20200

PR #404, opened 2024-09-12
[C-S] Make it use DCAwareRoundRobinPolicy unless rack is provided

fruch:
Currently c-s uses RackAwareRoundRobinPolicy in any case which leads it targeting particular rack only, even if rack is not provided.

This PR makes it use DCAwareRoundRobinPolicy unless rack is provided.

Ref: https://github.com/scylladb/cassandra-stress/issues/20
Ref: https://github.com/scylladb/cassandra-stress/pull/21


sylwiaszunejko (approved):

CodeLieutenant (approved):

denesb:
Submodule update: https://github.com/scylladb/scylladb/pull/20562.

